PR #558, opened 2015-02-16
Fix incorrect program output

olberger:
30 instead of 5


olberger:
Hope this helps

bnmnetp:
Thanks!

PR #564, opened 2015-02-22
Added missing posToNodeId method for Knight's Tour section of pythonds

acecodes:
This method's code was missing although it was being used by other methods on the same page.


acecodes:
You're right, that's a much better solution. This is what happens when I code all night on little sleep...

PR #581, opened 2015-04-17
Pip2

charbear:
change the word from lilteral to literal.

The evaluation of an expression produces a value, which is why expressions can appear on the right hand side of assignment statements. A lilteral all by itself is a simple expression, and so is a variable. Evaluating a variable gives the value that the variable refers to.

Run


bnmnetp:
Hi although your comment makes it appear as though this pull request is to fix a single typo, there are some 64 files modified.

Did you really mean to make a PR with all of those changes?  Could you give me a summary of what you changed?  Are a you a student working with Paul?

Brad

charbear:
So on one of the pages there is a typo so I posted the paragraph and they typo, I am one of Professor Resnicks students. I thought it would work that way, I just have to figure out how to specify the page. 

Sent from my iPhone

bnmnetp:
I think that maybe you meant to create an issue to report the typo which is great.  But in fact you have created a pull request.  The pull request contains changes to many files made by Paul and other students.

PR #586, opened 2015-04-30
New timed exam feature

ericsonga:
Added a new feature to allow for timed exams.  Modified the local
storage of answers to multiple choice questions and fill in the blank
questions to use the user id in front of the divid.


bnmnetp:
Hi Barb,

Could you modify overview/overview.rst to provide an example of how to use this?

I tried myself and got a little something going but I would like a section in overview to show everything in action.  What should the required parameters to start/stop timer be?  Should they be the same identifier to show that they go together or do they need to be unique?

Two problems:

1.  If I don't provide a duration paramter to startimer the build crashes
2.  The stoptimer stuff does not appear to be styled correctly.  Here's how it looks for me:



Notice how the finish button is way over on the right, and the statistics look like they should be in a highlight box or something.

Brad

ericsonga:
I added documentation to the overview and an example of how to use it.  I also centered the buttons and added a green highlight around the results.

PR #607, opened 2015-07-16
Error in prim's spanning tree algorithm listing

st4lk:
Look's like code for Prim's spanning tree algorithm is not correct.
Simple test. Check out this graph:

```
        A
       / \
    5 /   \ 9
     /     \
    B ----- C
        5
```

Correct minimum spanning tree must be (from node A):

```
        A
       /
    5 /
     /
    B ----- C
        5
```

whereas code from listing will produce following tree (same as dijkstra algorithm):

```
        A
       / \
    5 /   \ 9
     /     \
    B       C
```


bnmnetp:
Thanks for catching that.  That mistake has been around for a long time.

I will apply this change to the RunestoneInteractive/pythonds repository.

Brad

PR #609, opened 2015-07-23
small group improvements

mmoldavan:
- fixed the repeated first name on the group member list.
- fixed an exception that happens comparing a date type to a datetime type in the schedule view when plan end date is less than the current date.
- fixed the Continue Studying link under each chapter plan to go to the right book. It was hard coded to thinkcsp. 
- cleaned up some querying and added some checks that the user is logged in on mygroup methods.
- I added some filler text on how to get started if nothing is scheduled. right now it is a blank page when you first create a group.

### DIFF
--- a/controllers/mygroup.py
+++ b/controllers/mygroup.py
@@ -39,6 +39,7 @@ def schedule():
                                                                                 db.cohort_plan.cohort_id,
                                                                                 db.cohort_plan.id)
         dbfile.write('%s : %s\n' % (datetime.datetime.now(), 'before for plan'))
+
         for plan in allPlans:
             if plan.cohort_plan.status=='new' and plan.cohort_plan.start_date < datetime.datetime.now().date():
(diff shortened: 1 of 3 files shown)


bnmnetp:
I see that you modified some queries.  I have not studied them carefully myself, but please make sure you are thinking about efficiency.  I had to go back and update a bunch of queries related to small groups before that were super inefficient and blowing up my memory limits on webfaction.

mmoldavan:
Yes, the queries in here are confusing and there's room to optimize them. The query changes I did should be more efficient than what was there before. In the modify method I replaced 6 db updates to the same record with a single update. In the schedule method's allProgress query, I eliminated the inner join and return of the auth_user table, because it appears that table result was not being used anywhere. There is a query right after it to get all the users anyway.

On a lot of these joins I'm assuming the ID fields are indexed. Is there anywhere that lists which fields are indexed?

PR #634, opened 2015-08-29
removed requirement for subprocess32

presnick:
I found many instances of import subprocess (built-in python module from 2.6 onward). I didn't find any references to subprocess32. 

subprocess32 seems to be deprecated for windows. If it's not truly required, removing this requirement will save some headaches for people installing.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -10,6 +10,69 @@ Runestone Interactive Tools and Content
 
 Build Status |buildstatus|
 
+Relationship to other Runestone components
+------------------------------------------
+
+If you just want to use Runestone's capabilities to build pre-existing books or to make your own books using Sphinx and directives like ActiveCode, then you won't need this repository at all.
+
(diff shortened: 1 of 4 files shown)


bnmnetp:
It is most likely a dependency for one of the other packages.  pip does not distinguish.  But I haven't tried to remove it to see if something breaks.

presnick:
I haven't found a way to install subprocess32 on Windows 7, and some warnings that it's a bad idea, though I didn't completely undertsand the warnings. I will try to see if anything breaks when I run without it.

bnmnetp:
grepping through all of the .py files in site-packages of my virtual environment I find many that import subprocess but not subprocess32.  subprocess32 is a drop in replacement that is recommended for unixy systems.

My current thinking is that this was a dependency but that dependency was probably removed in a newer version of some package.

PR #645, opened 2015-09-29
Various fixes and features for grading, user bio, etc.

presnick:
I *think* common/js/bookfuncs.js has moved to RunestoneComponents. If so, let me know and I will git rm it.

Sorry that the installation documentation in README.rst isn't quite complete and I left all the old stuff at the bottom of the page. I think it's a useful start, though.

I think there may be a problem with ordering and timestamps in ajax.py. In ajax/getprog, I had changed, on line 180, to order by id (not timestamp) ascending and take the last one. That will work even if there is a null timestamp.

The source of null timestamps is our grading interface. If there is no existing saved code, then it creates an entry in the code table (with null timestamp) to save the grade score (presumably 0) and the comment if any. Your solution of filtering out the null timestamps in getprog is probably fine, too, and a little cleaner. We do need to fetch the rows with null timestamp for calculating overall grade on a problem set, but I don't think we need that in getprog(). That happens in assignment_get_scores() in grouped_assignments.py.


modules/luther/sphinx/datafile/__init__.py
```diff
@@ -0,0 +1,175 @@
+# Copyright (C) 2011  Bradley N. Miller
```
bnmnetp: This should not be a part of this PR as the datafile directive is part of RunestoneComponents

common/js/bookfuncs.js
```diff
@@ -185,6 +185,7 @@ function builtinRead(x) {
     return Sk.builtinFiles["files"][x];
```
bnmnetp: this copy of  bookfuncs.js is obsolete and should be removed.

controllers/ajax.py
```diff
@@ -176,7 +176,8 @@ def getprog():
         result = db(query)
         res['acid'] = acid
         if not result.isempty():
-            r = result.select(orderby=~codetbl.timestamp).first().code
+            # get the last code they saved; id order gets that for us
+            r = result.select(orderby=codetbl.id).last().code
```
bnmnetp: So I guess ordering by id is just as reliable as ordering by timestamp, but avoids the null problem.

presnick: Yes, exactly.

common/js/bookfuncs.js
presnick: Removed.

modules/luther/sphinx/datafile/__init__.py
presnick: Removed.

PR #650, opened 2015-10-04
keep user from changing their username

bnmnetp:
@presnick -- This is another way to prevent problems with users changing usernames.   Once the username is set, it cannot be changed.


presnick:
I like it. In my installation, I have removed the edit profile option entirely. But this is more elegant and will work for everyone.

PR #713, opened 2016-06-03
Move the chapternames module into Runestone Components

bnmnetp:
This move allows us to do full builds of books from the Jenkins CI machine including the necessary database updates for a book served by interactivepython.org


matt-oconnell:
@bnmnetp , this seems to cause an error:

`No module named runestone.server.chapternames` when running locally on web2py. Is there an additional configuration step necessary to run this successfully with `runestone build`?

PR #706, opened 2016-04-29
Teacher dashboard

mmoldavan:
Hey Brad, this is the teacher dashboard activity visualizations I built for my master's project geared towards Barbara's Computer Science Principles EBook over the past couple of months. There are many areas for improvement, but it may be a useful starting point for the teacher interface changes you mentioned and put some sketches up on google groups. I didn't add it to the navigation since it's not quite production ready.

Dashboard main page
project-final1b
Exercise analytics
project-final2
student report
project-final3b

PR #725, opened 2016-06-24
Admin tab

millro04:
Admin tab for the new instructors interface. The path to view it is 'runestone/admin/instructors.html'.

### DIFF
--- a/controllers/admin.py
+++ b/controllers/admin.py
@@ -3,6 +3,7 @@ import os
 import pygal
 from datetime import date, timedelta
 from paver.easy import sh
+import json
 
 
 # this is for admin links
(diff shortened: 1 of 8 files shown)


bnmnetp:
Wow!  This is great work.

I would really like it if a few of the @RunestoneInteractive/contributors  could check it out.  Its easy  From your runestone server master:

```
git checkout -b millro04-master master
git pull https://github.com/millro04/RunestoneServer.git master
```

Then start your server and go to `runestone/admin/instructors.html`

One thing we did not discuss in Ann Arbor was -- what to do with / how to publish an assignment!

I think we are missing a button that causes a bunch of rst to be written out to a file that can be incorporated into the book for the class.  But I would really like to hear what others think on this matter.

bnmnetp:
I've been giving this a lot of thought today as I've been looking at integrating #706 and this one together.  Here is what I think we need to do to make this work.

Each of the tabs should call its own controller.

The basic tabs structure should move to its own file.  Then the hrefs will not be #admin etc, but will be full URL's to the appropriate controller.  This will end up being a fairly small file.  For the sake of argument lets call it dashboard.html

The view for each individual controller will then extend dashboard.html and include everything needed for that particular tab.

* right now every database call needed for every tab is put into the instructor function.  This is OK for small test classes but will take forever for a large class. 
* Breaking it up into separate calls uses resources better.  We only fetch the data we need when an instructor actually wants to view that tab.
* This will make it much easier to integrate #706 into our look and feel
*  I think it will also make it easier to reuse a lot of the work that Paul did on the existing grading interface.

@millro04, @hangde01 lets do a hangout to discuss early next week, if you are available.

@ericsonga, @presnick I think we need to schedule a time to do a hangout and review the assignment tab, but more importantly to talk about how we take what is in the assignment tab and push it out to the the class.

PR #727, opened 2016-06-29
Github API  Bug Reports

hangde01:
I replaced the old problem page that directed the user to go to Github and report an issue there with a new page that, when the form is filled out and submitted, will automatically post a new issue into the the repository of the book that the problem originated in. This will require github credentials to be added into 'default.py'

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -28,6 +28,7 @@ pylint==1.2.1
 pyparsing==2.0.3
 python-dateutil==2.4.2
 pytz==2015.4
+requests==2.10.0
 runestone>=2.0.5
 six==1.9.0
 snowballstemmer==1.2.0
--- a/views/default/reportabug.html
+++ b/views/default/reportabug.html
@@ -2,7 +2,7 @@
 
 <h1 style="text-align: center">Report a Bug</h1>
 
-<p>Sorry you are having trouble. I want to hear about problems you are having but before I send you off to github to report a bug please look over the following list of common problems and solutions:</p>
+<p>Sorry you are having trouble. I want to hear about problems you are having but before send a bug report please look over the following list of common problems and solutions:</p>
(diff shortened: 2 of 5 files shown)


bnmnetp:
Also you are going to need to make a tiny little PR against the RunestoneComponents so that the link works from the static built page too.  It should be under runestone/common/project_template/_templates/....

hangde01:
I've been looking into using an access token since this morning, I just wanted to get what I had up. I'll keep looking and see if I can get that, since I agree it would likely be the easiest solution. I'll get the components file changed as well.

views/default/reportabug.html
```diff
@@ -0,0 +1,44 @@
...
+          <label for="coursename">Course Name</label>
+          <input type="text" class='form-control' placeholder="thinkcspy, pythonds, or other" name="coursename" id="coursename" required/>
+        </div>
+      </div>
+       <div class="form-group row">
+        <div class="col-md-4">
+          <label for="pagename">Page URL</label>
+          <input type="text" class='form-control' placeholder="Page URL of the bug" name="pagename" id="pagename" required/>
+        </div>
+      </div>
```
bnmnetp: I think you can autofill the values for course name and page url using information from the request object.  Especially since the course name is part of the url. I think it it will be in `referring page` or some other attribute of request.

bnmnetp:
@hangde01 instead of using the referer, you can use the template variable pagename, this should be a reliable way to pass both the pagename and the name of the course for a generated page.   http://www.sphinx-doc.org/en/stable/templating.html

requirements.txt
```diff
@@ -28,8 +28,12 @@ pylint==1.2.1
 pyparsing==2.0.3
 python-dateutil==2.4.2
 pytz==2015.4
+<<<<<<< HEAD
 requests==2.10.0
 runestone>=2.0.5
+=======
+runestone>=2.3.5
+>>>>>>> upstream/master
```
bnmnetp: Ooops, you checked in a version without resolving the conflicts.  I will fix this.

PR #747, opened 2016-07-18
This is a simple example of unit testing a controller.

bnmnetp:
Here is an example of doing unit testing of the web2py controllers.  I picked a pretty easy one to start with, but hopefully you will get the idea.

I guess I've been listening to too many "Talk Python to Me" episodes about testing on recent bike rides, but I have half a mind to not accept any more PRs that change a controller without the accompanying unit test...


bnmnetp:
Merged this manually to master.

PR #758, opened 2016-07-25
Dashboard

bnmnetp:
I wanted to rename the branch to something besides millro04-master to reduce confusion.  This is the same PR I created Saturday, but against the dashboard branch.

I've been giving this a lot of thought today as I've been looking at integrating #706 and this one together. Here is what I think we need to do to make this work.

Each of the tabs should call its own controller.

The basic tabs structure should move to its own file. Then the hrefs will not be #admin etc, but will be full URL's to the appropriate controller. This will end up being a fairly small file. For the sake of argument lets call it dashboard.html

The view for each individual controller will then extend dashboard.html and include everything needed for that particular tab.

right now every database call needed for every tab is put into the instructor function. This is OK for small test classes but will take forever for a large class.
Breaking it up into separate calls uses resources better. We only fetch the data we need when an instructor actually wants to view that tab.
This will make it much easier to integrate #706 into our look and feel
I think it will also make it easier to reuse a lot of the work that Paul did on the existing grading interface.
@millro04, @hangde01 lets do a hangout to discuss early next week, if you are available.

@ericsonga, @presnick I think we need to schedule a time to do a hangout and review the assignment tab, but more importantly to talk about how we take what is in the assignment tab and push it out to the the class.


presnick:
@bnmnetp @ericsonga I propose a hangout on Wednesday at 3:30PM eastern time.

bnmnetp:
@millro04 This new interface is really coming along nicely.

A couple of things we need to look at:

1. We need to allow the instructor to modify the due date, and/or the description after the assignment has been made.
2. The tabs on the left that indicate Formative / Summative / External AND the dropdown on the right hand side are confusing.  These should be in sync, or we should remove the one on the right hand side and only add to the section that is selected on the left.

I'm still trying to test the grading interface but I ran into problems with #1 causing my code to crash and not write out the assignment file.

3.  Lower priority as it is not used that often, we need to add another option in the admin tab to allow the user to edit their table of contents.  We already have an old interface for doing this, so it should not be hard.

bnmnetp:
BTW, my plan is to merge the dashboard branch later today or tomorrow and get this out on the live site.  We'll add a link to the new interface and label it BETA so that we can get some instructors helping us test it.

millro04:
Sounds good.

1 should be an easy addition to make. 

For 2, I think it'll be easiest to just remove the dropdown on the right. My original thought process with the dropdown was that they could choose to be viewing whatever type on the left but might not necessarily want to add to that same section. But it can easily be removed like you said. 

3. You're spot on about search,  I have never really implemented a search like that so I just iterated manually through everything. You will probably find the code to be messy/too complicated, but I just wanted to get a working version first. It is 'def questionBank' in admin.py. 

4. As of now I don't think you are even allowed to search for just summative questions, because the type is relative to your own assignment. So the questions table doesn't know whether it is summative, formative or external. Because one teacher can make a question summative while another teacher could make the same question formative if they choose. So the type is relative to the specific assignment. We do have the field question_type, which is activecode, etc. so maybe that is what we need to utlilize?

bnmnetp:
For 3, you can easily speed things up! You can build up a query pieces at a time. For example when you check for request.vars['author'] you can create `authorQ = db.questions.author == request.vars['author']`   Then later you can glue the final query together using the `&` operator  so if they specified both author and chapter then the query would end up `authorQ & chapterQ`  You can build up all the pieces into a variable and then use that as the parameter to your db call on line 918.

4.  You are right, for summative we can just search for question as the type -- another clause to add to your query above.

PR #779, opened 2016-07-28
Fixed issues from assignments tab initial commit

millro04:


millro04:
This should fix issues 771, 774 and 775.

PR #780, opened 2016-07-29
Make base course search criteria for q bank

millro04:
This should fix issue 773.

### DIFF
--- a/controllers/admin.py
+++ b/controllers/admin.py
@@ -595,8 +595,14 @@ def assignments():
     course_url = path.join('/',request.application, 'static', auth.user.course_name, 'index.html')
 
     print("ready")
+    row = db(db.courses.id == auth.user.course_id).select(db.courses.course_name, db.courses.base_course).first()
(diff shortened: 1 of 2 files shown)


bnmnetp:
OK, that looks fine.   Now, we could make it MUCH easier if we made the chapter a dropdown.  I think you could do this very easily because now that you know the base_course you can select the chapter labels with a simple select like this one:

```
select chapter_label from chapters where course_id = 'thinkcspy';
```

Can you add that to this PR?

millro04:
Yep, will course_id be 'thinkcspy' or the integer id corresponding to the course?

bnmnetp:
course_id is a string just like in the example

millro04:
Ok, see if that works. I didn't have any chapters in my table so I just made up a few for testing purposes, but I think it should work.

millro04:
This last commit makes the default option for the select to be nothing, that way we aren't forcing them to pick one of the chapters in the search if they don't want to search for that.

PR #795, opened 2016-08-07
added check to ensure filled entries

hangde01:
Added a JavaScript function to ensure that relevant fields are filled in.


bnmnetp:
Looks good.  While you are at it can you clean up the email and username thing.  They should be in separate form inputs.  And one or the other should be required.  anonymous bug reports are pretty useless, so go ahead and autofill those fields like I'm doing in the detail page now, and take them out of the detail.

PR #812, opened 2016-08-15
Displayed course name and fixed rebuild issue

millro04:
Let me know if you don't like where the course name is displayed/how big it is. This should also reload the page to the rebuild tab after clicking rebuild.


millro04:
With this last commit you can now edit index.rst from the admin tab. I also fixed an issue I found with creating/editing questions, where if there were more than two new line characters, one of our  regular expressions only replaced it with one new line, so that was also fixed.

PR #837, opened 2016-08-25
Update build process checking

presnick:
I was getting errors due to file system permissions (scheduler running with different permissions that webserver, for security reasons). There was some outdated code that was writing a "done" file and using that to see if the process was done. There was also some code to check the scheduler process status in the database. I switched it to use just that code, and updated the javascript to handle the new return codes. 

Also changed the *old* build and rebuild pages to give a little more diagnostics about scheduler being missing if the process never starts. If you'd like to have that in the new interface's build, let me know and we can easily add it to views/admin/admin.html. I'm not quite sure about integrating it with dd.js, though.


bnmnetp:
I would definitely like to see this in the new course as well.

I need to do a bit more checking, but I think that dd.js can probably be deleted.  As far as I can recall it was only used by the old drag-and-drop interface for arranging the chapters and we have not used that in a couple years.

Do you know when the status value changed?

presnick:
Looks like designer/build.html has a Create Course button that invokes buildIndexFile, which invokes the code in dd.js.

But that code also has inline jscript to do the same checking, I think. I'm not quite sure what's happening there, which is why I'm reluctant to get rid of that change in dd.js.

Not sure what your question is about when the status value changed. I changed getSphinxBuildStatus to just return the status code from the scheduler database rather than translating it into our own words for the same codes.

bnmnetp:
buildIndexFile is referenced in the else clause that will never get activated in build.html

That should be cleaned up and dd.js removed.

presnick:
Doing the cleanup now.

presnick:
Cleaned up. Should be ready for merge now.

presnick:
One more commit added, for adding gradeable_div field to questions table, as part of resolution of https://github.com/RunestoneInteractive/RunestoneComponents/issues/248

PR #842, opened 2016-08-27
Clean autograde

presnick:
(Has previous pull request from master, plus...)

Stuff for autograding.


models/questions.py
```diff
@@ -19,6 +21,18 @@
 except:
     pass
 
+db.define_table('question_grades',
```
bnmnetp: @presnick This needs to be moved to the end of db_ebook.py   Both Jackie and I ran into problems until I figured out that we needed to relocate the definition of this.

I don't have a good explanation for why, but something in the ordering of the model loading was causing big trouble.  Moving it fixes this.

presnick:
This pull request is now superseded by a PR from a different branch, with additional stuff and fixes.

The offending table definition was moved to the end of grouped_assignments.py; hopefully it won't cause migration problems there, and that's where it belongs conceptually, with other grading-related stuff.

PR #846, opened 2016-09-01
Autograding

presnick:
This branch also includes changes that were pushed in https://github.com/RunestoneInteractive/RunestoneServer/pull/837

This PR supersedes the PR in https://github.com/RunestoneInteractive/RunestoneServer/pull/842, which should not be used.


bnmnetp:
Paul,  So, I interpret this to mean that I can close PR's #837 and #842 to remove any possible confusion?

presnick:
Yes.

sent from mobile-- strange words are probably auto-correct gone awry.

Issue #851, opened 2016-09-05
get_problem crashing in grading interface


bnmnetp:
The `get_problem` function has problems.  At least in the old grading interface...

It is looking for request.vars.assignment when there is none .. I think it is looking for id  This results in the assignment begin None and things go downhill from there.

bnmnetp:
I made a quick and dirty patch which seems to keep things going, but not sure.

```
    assignment_name = request.vars.assignment
    if not assignment_name:
        assignment_name = request.vars.id

    assignment = db((db.assignments.name == assignment_name) & (db.assignments.course == auth.user.course_id)).select().first()
    try:
        deadline = assignment.duedate
    except:
        deadline = None
```

presnick:
Will look into this tomorrow.

sent from mobile-- strange words are probably auto-correct gone awry.

PR #853, opened 2016-09-05
fix #852 

### LINKED ISSUE #852: old grading interface broken; assignment/detail.html
Crashing because of gradeRecordingURL not found.

presnick:
gradeRecordingURL not needed for old grading interface

### DIFF
--- a/controllers/admin.py
+++ b/controllers/admin.py
@@ -711,8 +711,7 @@ def grading():
         for chapter_q in chapter_questions:
             q_list.append(chapter_q.name)
         chapter_labels[row.chapter_label] = q_list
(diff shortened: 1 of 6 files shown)


presnick:
This PR now fixes https://github.com/RunestoneInteractive/RunestoneServer/issues/851 as well. The safe fix for this is to leave def problem() completely unchanged, as that's an endpoint that the old grading interface uses, and make the new grading interface use comletely new endpoints.

I was trying to be clever. Since I created one of the new endpoints, get_problem() by copying some of the code from the old problem(), I thought I'd refactor the old problem() to do call the new endpoint. This was a bad idea because I subsequently edited get_problem to handle something needed in the new grading interface, which broke the old grading interface. Who knows what else we might add to get_problem in the future. So I'm just breaking the dependency and letting the old code for problem() live as it was. Eventually, we will remove the old grading interface and, if we can remember, we can remove def problem() entirely at that point.

PR #855, opened 2016-09-05
Add some space on the edges of the page

mlhess:
The text runs into the borderwall


bnmnetp:
What page was giving you trouble??  Usually I hear the opposite.

PR #869, opened 2016-09-18
autograde type comes from assignment_questions; getassignmentgrade looks in new location first

presnick:


models/questions.py
```diff
@@ -15,10 +15,10 @@
...
 # In SQL we can manually add the constraint
 # alter table questions add constraint name_bc_unique UNIQUE(name, base_course);
-try:
-    db.executesql('''alter table questions add constraint name_bc_unique UNIQUE(name, base_course)''')
-except:
-    pass
+# try:
+#     db.executesql('''alter table questions add constraint name_bc_unique UNIQUE(name, base_course)''')
+# except:
+#     pass
```
bnmnetp: If this is commented out, then new installations will not get this constraint which we want to preseve?  Why comment it out?

controllers/assignments.py
```diff
@@ -529,13 +528,13 @@ def _autograde_one_q(course_name, assignment_id, sid, qname, points, deadline=No
     question = db(db.questions.name == qname).select().first()
 
     # dispatch on grading_type; if none specified, can't autograde
-    if question.autograde == 'unittest':
+    if autograde == 'unittest':
```
bnmnetp: If an instructor is following a grading path of chapter --> question will this still work?  Why the change from autograde to question.autograde ??

controllers/ajax.py
```diff
@@ -707,11 +707,12 @@ def getSphinxBuildStatus():
...
             results['status'] = row['status']
         else:  # task failed
-            results['status'] = 'failed'
+            results['status'] = row['status']
             tb = db(db.scheduler_run.task_id == row.id).select().first()['traceback']
             results['traceback']=tb
     else:
         results['status'] = 'failed'
+        results['info'] = 'no row'
     return json.dumps(results)
```
bnmnetp: This looks like it is unrelated to the purpose of the PR -- but seems like valid cleanup??

models/questions.py
presnick: I was getting an error when there is another migration to be done later in the file. The attempt to run db.executesql seems to poison the session objec that will be needed later, so try/except wasn't sufficient. (Also seems inefficient to keep running this query every time web2py loads. Isn't there a way to get this constraint through web2py define_table?)

controllers/assignments.py
presnick: I wasn't aware of that grading path. I don't think I implemented autograde for anything other than assignments. What do you have in mind?

Note: we discussed last week and decided the decision about whether particular questions should be autograded should be a property of the assignment_question, not the question.

controllers/ajax.py
presnick: Yeah, I must have encountered some issue where I needed that for debugging purposes.

models/questions.py
bnmnetp: According to everything I could find in the web2py docs, you cannot create a constraint like this as part of the regular syntax for defining a table.

controllers/assignments.py
bnmnetp: We talked about it this summer.  The path is where the instructor selects Chapter for the first column, a question from the chapter for the second, and then student in the third column.

presnick: OK. But I think that's just not implemented at this time. I suggest that we make that a to-do item. (Unless you think it's actually working and this will break it.)

models/questions.py
presnick: Hmm, well we need to do something other than this, because it breaks things when there's another migration to be done.

controllers/assignments.py
bnmnetp: I think I may have been confused about the recent conversation or maybe I'm confused by the values associated with the autograde column.

1.  Whether or not a question can be autograded is one thing.  I can see where this would be useful at the assignment level.  If you try to autograde an entire assignment there might be questions that are autogradeable and questions that are not.   But, if a question is autogradeable at all isn't it always autogradeable?

2.  How a question is autograded seems like a separate idea, and does not seem likely to change as the autograde method is kind of encompassed by the kind of question it is in the first place.

PR #874, opened 2016-09-23
Grading fixes

presnick:


presnick:
This has some important bug fixes since the last PR, but we'll have to resolve that one first.

bnmnetp:
OK, I accepted that last one, but this one now has conflicts with your previous PR

presnick:
Merge conflict resolved.

controllers/ajax.py
```diff
@@ -219,24 +219,23 @@ def gethist():
...
-    sid = request.vars.sid
 
-    if sid:
-        query = ((codetbl.sid == sid) & (codetbl.acid == acid) & (codetbl.timestamp != None))
+    if request.vars.sid:
+        sid = request.vars.sid
+    elif auth.user.username:
+        sid = auth.user.username
     else:
-        if auth.user:
```
bnmnetp: I don't think this should be removed.   If the sid is a parameter then we are getting the history for that sid from the grading interface.  But if a student is logged in then we get the sid from their logged in username.  How will this work if this is deleted?

static/js/admin.js
```diff
@@ -894,7 +906,7 @@ function getStudents(sectionName) {
     obj.onreadystatechange = function () {
 
         if (obj.readyState == 4 && obj.status == 200) {
-            students = JSON.parse(obj.responseText);
+            var students = JSON.parse(obj.responseText);
```
bnmnetp: Where are all of the uses of students above getting their value, especially if this is declared local, and you have removed all the other calls to JSON.parse.

Issue #911, opened 2017-06-07
unsupported operand type(s) for +: 'int' and 'NoneType'


bnmnetp:
```
Traceback (most recent call last): 
File "/home/bnmnetp/webapps/uwsgi_runestone/web2py/gluon/restricted.py", line 227, in restricted 
exec ccode in environment 
File "/home/bnmnetp/webapps/uwsgi_runestone/web2py/applications/runestone/views/assignments/index.html", line 125, in <module> 
File "/home/bnmnetp/webapps/uwsgi_runestone/web2py/applications/runestone/models/grouped_assignments.py", line 104, in points 
return sum([t.points(projected, potential) or 0 for t in self.assignment_type_grades]) 
File "/home/bnmnetp/webapps/uwsgi_runestone/web2py/applications/runestone/models/grouped_assignments.py", line 86, in points 
return sum(vals[self.assignments_dropped:]) 
TypeError: unsupported operand type(s) for +: 'int' and 'NoneType'
```

@conzty01 and @gardhi01 :  This will be a good one to get you looking at the server side.

PR #912, opened 2017-06-07
Protected against adding an int and None

conzty01:
The issue became that the db could return a None or null value for an assignment score and would result in summing an int and NoneType down the line.  Adding a try/except catches datatype errors and replaces them with 0s.

### DIFF
--- a/models/grouped_assignments.py
+++ b/models/grouped_assignments.py
@@ -9,7 +9,7 @@ class AssignmentGrade(object):
         self.released = released
         self.score = score
         self.projected = projected
-        try: assignment_score:
+        try:
             self.assignment_score = int(assignment_score)
         except:
             self.assignment_score = 0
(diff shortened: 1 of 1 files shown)


models/grouped_assignments.py
```diff
@@ -8,13 +8,16 @@ class AssignmentGrade(object):
     def __init__(self, released, score, projected, assignment_score, assignment_id, assignment_name, grade_record, row):
         self.released = released
         self.score = score
-        self.projected = projected 
-        self.assignment_score = assignment_score
+        self.projected = projected
+        try: assignment_score:
```
bnmnetp: This line looks like a syntax error.  `try: foo:`  the `foo:` part is definitely wrong.

bnmnetp:
That looks good.    Thanks!

In the future, when you make a PR it would be great if you could reference the number of the issue in the PR.  If you say fixes #911 then you will be able to click on the number to see the issue, AND the issue will automatically be closed when I accept the PR.

Brad

Issue #933, opened 2017-06-23
Automatically calculate totals in admin/grading


presnick:
Whenever a new grade is sent to the DB, also call calculate totals.

Remove the special button for that.

conzty01:
@conzty01 @gardhi01

Issue #895, opened 2017-03-10
Summative assignments should have null threshold in DB


bnmnetp:

gardhi01:
@conzty01

PR #936, opened 2017-06-23
Updated Exercise Analytics in Instructor Dashboard

conzty01:
Questions under the "Exercise Analytics" section of the Instructor Dashboard now link to corresponding pages in the course book so the instructor can see the question.


conzty01:
Linking #895 #933

conzty01:
I don't know if having all new assignments default to null is the way to go, but I can't see anything in there currently that depends on the threshold being set aside from the grading which will give the students full credit if the student scores higher than the threshold. This is an inconvenience when `threshold = 1` as the instructor will have to go back and change the points. I also know from slack that Paul Resnick is working in this area which may cause a merge error, but this should be a decent temporary solution.

Issue #940, opened 2017-06-27
Broken admin interface


bjones1:
In getting runestone up and running on my work PC, I built everything from scratch. From what I can tell, there's no longer a way to get to the admin interface for a newly-created user -- `runestone.controllers.admin.index` (which only required a login) redirects to `runestone.controllers.admin.admin` (which calls `verifyInstructorStatus`). I think I need a way to get to `runestone.designer.index` to create a new course which I own, but there's no way to this page except through the admin interface, which requires me to be a designer.

Or did I overlook something obvious?

bnmnetp:
You beat me to it, I was going to do a clean install to a linux VM today.

This is definitely a flaw in our logic and testing.  All of us testing already had users that were instructors.

I think we need a big giant button somewhere on the front page that says "I WANT TO MAKE MY OWN COURSE AND BE THE INSTRUCTOR"  This button would take you to the designer.  It is one of the most FAQs I get by email.

bjones1:
:) Sounds good to me...

bnmnetp:
Add button to the home page that takes them to designer/index controller.

Issue #934, opened 2017-06-23
Always show assignment total for currently selected student


presnick:
On admin/grading, whenever there is a currently selected student, show the form  with total points and the ability for instructor to override with a manual total.

That form is currently shown right after the calculate totals button is pressed.

conzty01:
@conzty01 @gardhi01

PR #943, opened 2017-06-28
Added button to designer/index from home page (Issue #940)

gardhi01:


gardhi01:
The commits that display the assignmentTotalsForm are linked with #934

gardhi01:
I found that rebuilding a course on admin/admin did not have a course_id defined and would therefore not allow a course to be rebuilt. I added the course_id to the course rebuild return statement.

Issue #906, opened 2017-05-22
Instructor Help link


psmaker:
I get a 404 Not Found when pressing the Instructor Help link on the upper right hand corner of the main page.

http://interactivepython.org/runestone/static/overview/instructor.html

PR #950, opened 2017-07-05
fixed instructor help link for home page and instructor pages to solve issue #906

gardhi01:
I saw there was another issue similar to this one that has been closed, but I was still unable to reach the help page from the instructor help links in the home page or the instructor interface pages. (The help link did work from the table of contents page though.)


bnmnetp:
Good catch Hillary, the books and the pages use two different templates.

Issue #924, opened 2017-06-20
autograde more practice types like Parsons and fill in the blank


bnmnetp:

bnmnetp:
Paul,

If you dont' think you can realistically work on this one, we could assign it to @gardhi01 and @conzty01 .


It seems to me this should be separable from the interface work you are doing.  right??

presnick:
Yes, please reassign it. 

The one point of connection with the interface is that @bjones1 will need to know the new strings that are allowed to designate the new type of autograding.

Note for @gardhi01 and @conzty01: the place to add more autograders is in assignments.py _autograde_one_q, where it dispatches on the autograde field's value: unittest, first_answer, etc. You will just define some new possible values for that field (e.g., for parson's problems,  "correct order", "correct order and indentation").

bnmnetp:
OK, maybe we can agree on those strings up front, and document what is already there here.

activecode:    unittest
mchoice: first_answer, last_answer
clickableArea
fillintheblank
dragndrop
parsons
timed_exam??
ALL: visited  ??

Hmm, it seems like first_answer, last_answer would be appropriate for clickableArea, fillintheblank, dragndrop and parsons.  But that is not unique to each type.  However the question type is in the database in the question table which looks like it is available in autograde_one_q so I think we can dispatch based on both pieces of information when needed.?

looping @bjones1 into the discussion.

bjones1:
Codelens problems can optionally provide a correct answer, right? So they can be autograded?

Also, since activecode problems can include a unit test, can't they also be graded as first_answer, last_answer, or does the database not store that?

presnick:
To be more complete, activecode could be:
First_answer
Last-answer
Last_answer_before_deadline
Best_answer
Best_answer_before_deadline

Also, unittests could be graded all_or_none, as they are now, or
percentage_of_tests_passed.

Somehow have to cross the choice of which code run to grade with the all vs
pct choice.

sent from mobile-- strange words are probably auto-correct gone awry.

bnmnetp:
Seems to me like we have several orthogonal concepts:

1.  Before deadline versus last answer before I grade  -- this is a filter on what questions answers could be considered.  My guess is that for most instructors this is a course policy.
2. first_answer, last_answer, or best_answer  -- this is a more specific filter on which of the available answers to actually grade.
3. unittest:  all_or_nothing versus percent_correct  -- this is a scoring algorithm

presnick:
[...]

PR #952, opened 2017-07-06
Working on making parsons and fill in the blank autogradeable for #924

gardhi01:


gardhi01:
Sorry about all the indentation changes--Sublime was formatting things strangely, so whenever I would add lines of code, I ran into indentation errors.
The helper function bit is there at Paul's request.

bnmnetp:
You need to tell sublime to indent by 4 spaces

gardhi01:
The autograder looks for first_answer and last_answer in autograde_one_q (around line 539). That's something that Paul already implemented. The elif statements that I added in autograde_one_q just collect all the answers that a particular student gave for a particular question and determine the scoring function.  Then the answer to be scored is determined by Paul's first_answer last_answer conditional.

bnmnetp:
OK, so....  the following should be autogradeable.  You can check off the one's that are complete.

- [ ] multiple choice
- [ ] activecode  -- using unit tests both all or nothing or a percentage of tests passed
- [ ] clickableArea
- [ ] dragndrop
- [ ] fill in the blank
- [ ] parsons
- [ ] usage -- @presnick may have this completed??
- [ ] codelens questions

gardhi01:
I'm not seeing clickableArea or dragndrop in the question_type column of the questions table. Where would I find those types of questions?

bnmnetp:
There is a database table for dragndrop_answers and clickablearea_answers just like the others.

They are newer and not as widely used in the textbooks yet.  And maybe there is missing code in their implementation that they are not adding themselves to the questions table like they should.

I'll investigate that side.

gardhi01:
Okay, I'll try to create my own questions and see how that affects things.

One other question I have is about the codelens questions.  I saw that Bryan said they could provide a correct answer, but codelens doesn't have its own answers table in the database. Would we want to create a codelens_answers table?

bnmnetp:
You can also clone the RunestoneInteractive/overview  and build that.  It has examples of both.

I think we do want to add a codlens answers table.

gardhi01:
What does the minheight column in the dragndrop_answers keep track of?

gardhi01:
I am slightly confused by codelens questions. I've looked at several, but there doesn't seem to be a way for a student to alter the code in any way--the only way a student can interact with the codelens question is by stepping through the provided code with the buttons or using the slider. How can a student be correct or incorrect with a codelens question?  Or is codelens one of those question types that should just be graded on a student's interactivity?

bnmnetp:
If you look at the overview you will see that there is an option to ADD questions to codelens so that when the student is stepping through the code it will pop up a modal and ask the student to predict either the value of a variable at that point, or what the next line to be executed will be.

bnmnetp:
The assignment pages are looking good @gardhi01 !

I've pulled everything onto the assignment_interface branch.  This includes the latest from Bryan and Paul for doing the readings part of the assignment.  So perhaps you could pull that branch to see how the readings are stored and add them to the assignment page too.??

gardhi01:
@bnmnetp The last commit I made will allow scripts inside the rendered html to run. This will let codelens questions render correctly within the dynamic assignment pages, but I'm worried it may be a bit of a security risk. What are your thoughts?

PR #954, opened 2017-07-07
Fixed issue with registering for a course

conzty01:


controllers/admin.py
```diff
@@ -351,8 +351,10 @@ def admin():
 
     return dict(sectionInfo=sectionsList, startDate=date.isoformat(), coursename=auth.user.course_name,
                 instructors=instructordict, students=studentdict, confirm=False,
+
```
bnmnetp: We don't want a blank line in the middle of a parameter list.

conzty01: It should be fixed @bnmnetp

PR #963, opened 2017-07-12
Fix: Error on Windows during a scheduled build.

bjones1:
...for that poor, little single-tasking origins of Windows...since it can't delete files that are open...


bjones1:
I just updated this to resolve merge conflicts. Would you take a look when you have a minute?

bjones1:
Thanks!

PR #965, opened 2017-07-14
Updates for New Timed Exams Functionality

conzty01:


controllers/ajax.py
```diff
@@ -911,11 +911,11 @@ def getAssessResults():
         res = {'answer': rows[0][0], 'timestamp': str(rows[0][1]), 'correct': rows[0][2]}
         return json.dumps(res)
     elif event == "timedExam":
-        query = "select correct, incorrect, skipped, time_taken, timestamp from timed_exam where div_id=%s and course_name=%s and sid=%s order by timestamp desc"
+        query = "SELECT correct, incorrect, skipped, time_taken, timestamp, reset FROM timed_exam WHERE div_id=%s AND course_name=%s AND sid=%s ORDER BY timestamp DESC"
```
bnmnetp: Its a minor thing, but it would definitely be preferable to use the `db((db.timed_exam.div_id == divId)` style query rather than using executesql.  The results are much more readable.

conzty01: All of the queries in this block of `if/elif` statements use the `db.executesql()` function. Would you like me to change them all to the web2py query style?

bnmnetp: That would be great.

controllers/ajax.py
```diff
@@ -883,43 +883,37 @@ def getAssessResults():
 
     # Identify the correct event and query the database so we can load it from the server
     if event == "fillb":
-        query = "select answer, timestamp, correct from fitb_answers where div_id=%s and course_name=%s and sid=%s order by timestamp desc"
-        rows = db.executesql(query, (div_id, course, sid))
+        rows = db((db.fitb_answers.div_id == div_id) & (db.fitb_answers.course_name == course) & (db.fitb_answers.sid == sid)).select(db.fitb_answers.answer, db.fitb_answers.timestamp, db.fitb_answers.correct, orderby=~db.fitb_answers.timestamp)
         if len(rows) == 0:
             return ""   # server doesn't have it so we load from local storage instead
         res = {'answer': rows[0][0], 'timestamp': str(rows[0][1]), 'correct': rows[0][2]}
```
bnmnetp: You need to update the use of rows -- this is the real benefit!  So this line can now be:

```
res = {'answer': rows.answer, 'timestamp': str(rows.timestamp), ...
```

Otherwise ajax.py will crash on line 889 and others.

conzty01:
Referenced Issue: #967

Issue #957, opened 2017-07-07
Adding reading sections to an assignment


bnmnetp:
When I select a chapter/subchapter from the tree-view I expect that it would add something to the assignment table to indicate that reading the section is part of the assignment.  But it does not appear.

Also, we discussed that we need to have some kind of flag to indicate whether a particular question should be reproduced on the assignment page, or left to be done in-line.  I expect that would be a column in the assignment_questions table, but I don't see it there yet?  Or perhaps there is some other way to know.  This will be important for #955 to work properly.

bjones1:
Fixed in https://github.com/RunestoneInteractive/RunestoneServer/tree/assignment_interface, please close.

bnmnetp:
OK, just a few comments about some polish that I think may help, and some questions.

1.  the divisions between whole assignment, reading part, and exercise part are not clear. especially where the whole assignment and reading part come together.   -- I think a clear delineation and even a simple heading would help. 
2.  There is lots of blank space to the right where some simple help text would guide instructors through the process. -- Remember many of our users are high school teachers that are not developers.
1. I'm assuming the autograder dropdown refers to how the reading assignment is graded, but I couldn't explain how that works.
1. I would have expected that the in-reading components such as multiple choice etc would have appeared under the reading section, but I guess I have to add them in the next section?  I think this will be counter intuitive for many instructors that just want their students to read these sections and do the widgets in the section.   In fact to be honest I was surprised that there were two different tree views, one for reading stuff and one for exercise stuff.
1. I think the Write Assignment button can go now, as Hillary will have a working dynamic assignment page generator for us shortly.
1.  It looks like the tree views are being generated by querying chapters and subchapters of the base course rather than the specific course. This is fine for everyone except those that are doing highly customized books.

bjones1:
Thanks, that's helpful. So I think:

- Items 1-3 are probably GUI improvements / explanatory text, correct? Perhaps item 4 as well?
- Item 5: Yay! I'll be happy to get rid of this gunk.
- Item 6: thanks for checking. I'll let @presnick fix this one.

bnmnetp:
Thanks for the update @bjones1, I'm afraid we had a big misunderstanding about the **Write Assignment** button.

[...]

PR #969, opened 2017-07-25
Assignment interface

bjones1:
Fixes as requested in #957.


static/js/admin.js
```diff
@@ -1295,117 +1301,6 @@ function remove_question(question_name) {
...
-        });
-    };
-
-    var hiddenwrite = document.getElementById('hiddenwrite');
-    hiddenwrite.style.visibility = 'visible';
-}
-
-
-// Called when the "Done" button of the "Write" dialog is clicked.
-function create_question(formdata) {
```
bnmnetp: This seems a bit overzealous.  We still want people to be able to write their own questions, preview them, and have them added to the DB.

views/admin/assignments.html
```diff
@@ -33,15 +33,31 @@
...
-<div class="col-md-6">
-  <h4 style="text-align: center;">Assignment</h4>
-  <select class="form-control" id="assignlist" onchange="assignmentInfo();">
-    {{ for assignmentid in assignments: }}
-    <option value="{{=assignmentid}}">{{=assignments[assignmentid]}}</option>
-    {{ pass }}
-  </select>
-  <button class="btn" data-target="#addAssignment" data-toggle="modal" style="margin-top: 12px;">+</button>
-  <button class="btn" onclick="remove_assignment();" style="margin-top: 12px;">-</button>
+<div class="row">
```
bnmnetp: I'll study these a bit more, but as I look at the interface, it seems that the ability to write your own questions for an assignment is completely gone.  Why?

bnmnetp:
OK, maybe there was confusion about what I meant when I said Write Assignment.  I meant that there is no need for the button in this screenshot to exist anymore.  But we definitely still want instructors to be able to write and add questions to the database.

screen shot 2017-07-25 at 12 11 14 pm

bjones1:
Oops! So, am I correct that I should:

- [ ] Revert the commit that removed the "Write" button.
- [ ] Remove the "Write Assignment" button and all the code behind it.

bnmnetp:
Yes, that is correct.

In the new world (which you should like!) The assignment page is generated dyamically on the server side.  Removing the Write Asssignment button will remove the code that updates the assignments.rst file and creates a new assignment in the assignments folder.  --- YAY! NO NEED to REBUILD your course just because you made an assignment.   Once you've restored the write button, questions written and added to assignments will work too!


I'll have to remember this, because this is a great example of why you should do good feature based commits !

bjones1:
OK, this part should be fixed.

PR #1004, opened 2017-08-10
Added ability to remove students from courses

conzty01:


conzty01:
Instructors can remove a single student or multiple.  There is also built in safety measures that will move the user to the base course and will not through an error if the "--Select Student(s) to Remove--" option is selection.  Additionally, if the instructor is selected, they will not be removed from the course.

conzty01:
I can foresee an issue in which the base course is not running on the server but I didn't know if the appropriate action would be to do nothing or to completely remove the student from being enrolled in any course. Thoughts?

bnmnetp:
Good work.  the base course has to be on the server in order for a course to be based on it.  So I don't think that will be a problem.

I made one minor change -- when I merged.  I updated the auth_user table to reflect the "current enrollment" for the student, and I added a flash message so the instructor would see that students were removed successfully.

PR #1009, opened 2017-08-13
alter admonition background

haugrud:
make the background a little more noticeable


bnmnetp:
I prefer to keep the background the more neutral tan color.  The light green conflicts with our usage of that color in many other places to indicate success.

PR #1032, opened 2017-08-30
Fix: Add questions to the question picker tree as needed.

### LINKED ISSUE #433: Add progress graphs to the instructor dashboard
show the total number of activities as a bar graph for the last 10 days on the dashboard page.  Also show student activity counts as well as event activity counts for a course.

bjones1:
Closes https://github.com/RunestoneInteractive/RunestoneComponents/issues/433.

### DIFF
--- a/controllers/admin.py
+++ b/controllers/admin.py
@@ -250,6 +250,7 @@ def assignments():
                 tags=tags,
                 chapters=chapter_labels,
                 toc=_get_toc_and_questions(),
+                chapterMap=_get_chapterMap(),
                 )
 
 @auth.requires(lambda: verifyInstructorStatus(auth.user.course_name, auth.user), requires_login=True)
@@ -481,7 +482,7 @@ def removeStudents():
             db((db.user_courses.user_id == int(studentID)) & (db.user_courses.course_id == auth.user.course_id)).delete()
             db.user_courses.insert(user_id=int(studentID), course_id=baseCourseID)
             db(db.auth_user.id == int(studentID)).update(course_id=baseCourseID, course_name=baseCourseName)
-    
+
     session.flash = T("You have successfully removed students")
     return redirect('/%s/admin/admin' % (request.application))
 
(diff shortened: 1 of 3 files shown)


bnmnetp:
I think I need you to walk me through some of the changes.    After all the time I've spent fixing bugs, I get nervous about big changes like this.

bjones1:
There were lots of big changes. I'd glad you asked for a walkthrough, because that gave me some time to reflect on what I wrote. I based all my changes on the assumption that the search interface can find things not in the question picker tree. Is that correct? If not, then a much smaller change might fix the problem.

bjones1:
As an alternative, if #1035 fixes the bug -- it assumes that the search interface always finds items that are already in the question picker tree -- then you can merge that instead. It's much simpler.

bnmnetp:
Yes, the search interface can potentially find things that are not in the tree.  Although I think with recent changes most, if not all, things are ending up in the tree.

bjones1:
So, would you like to go this route, or just merge #1035 as a partial fix for now? If you'd like to go this route, then I can provide the walkthrough of the changes.

bnmnetp:
When I pull this into my dev system it definitely breaks things that are working in master.  Particularly when you write a new question it no longer gets added to the table when you save the new question.   Also I had added defaults of 'manual' and 'last_answer' so that there would not be Blank showing on in the table for how to grade and what to grade.

bjones1:
Oops. I'll take a look and fix, then do a bit more testing.

So, I take it you'd like to pursue this PR, rather than #1035, and would also like a walkthrough to explain the large amount of risky changes?

bjones1:
I can't reproduce the "when you write a new question it no longer gets added to the table", but I did get questions double-added to the table. That's fixed now. I can search, write, and delete problems with no errors, as far as I can tell.

However, this brings up a related side-topic: both the Search and Write buttons offer a Points input and a Timed checkbox. IMHO these should be deleted -- points should be set in the table. Timed makes no sense anymore. Please confirm; I'm currently ignoring these, but haven't deleted anything.

Re: defaults, this really needs to be solved on the server. Current code has defaults of `pct_correct` for auto-grade and `best_answer` for what to grade, but this can't apply to all question types. Instead, client side code should send an undefined to the server, which should then pick the correct default based on the question type.  But this fix really belongs in a separate PR.

bjones1:
Old, and closing.

Issue #1027, opened 2017-08-28
Feature: Sort Names on Admin tab


psmaker:
It would be nice to have a sorted list (by last name) of names for both the Manage Section and Manage Students

PR #1033, opened 2017-08-31
Sort student names on grading page

shadaj:
Partial fix towards #1027

### DIFF
--- a/static/js/admin.js
+++ b/static/js/admin.js
@@ -523,12 +523,13 @@ function pickedStudents(column) {
     var studentslist = students;
     var keys = [];
     var i;
-    for (i in studentslist) {
+    for (i in list) {
         if (studentslist.hasOwnProperty(i)) {
             keys.push(i);
         }
     }
 
+    keys.sort();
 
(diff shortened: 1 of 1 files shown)


bnmnetp:
This makes the names sorted on the manage sections tab.  Which is a good thing!

The grading page is a bit trickier, I think, but the code for that is in the `pickedStudents` function

psmaker:
Just checking if it this pull request is OK. It would be highly useful for us with about 120 students, it takes a bit of time to find students' name.

static/js/admin.js
```diff
@@ -523,12 +523,13 @@ function pickedStudents(column) {
     var studentslist = students;
     var keys = [];
     var i;
-    for (i in studentslist) {
+    for (i in list) {
```
bnmnetp: Where does list come from??

bnmnetp:
@psmaker I don't think this is correct.  See my comment above??

bnmnetp:
One use case for the sections list that has emerged is when a student registers twice!  It would be nice to have the username displayed along with the First and Last name to disambiguate questions like that.

PR #1035, opened 2017-08-31
Fix: Check any questions found in the search interface.

bjones1:
Might fix https://github.com/RunestoneInteractive/RunestoneComponents/issues/433; simpler than #1032.


bjones1:
This is so old, I don't even remember what it's doing. Obviously, it's not bothering anyone.

PR #1061, opened 2017-11-22
Added checkfile proxy function for jobe server

CABowers:


controllers/proxy.py
```diff
@@ -5,21 +5,24 @@
 logger.setLevel(settings.log_level)
 
 
+response.headers['Access-Control-Allow-Origin'] = '*'
```
bnmnetp: This opens us up to anyone using this to proxy JOBE requests rather than only working for requests served by the runestone server.  Is there a use case for opening it up like this??

PR #1063, opened 2017-12-21
updates to do LTI insertion of assignment grades into LMS gradebook

presnick:
Format of URL for deep linking to an assignment from an LTI-enabled LMS:
https://sitedoma.in/runestone/lti?assignment_id=73


bnmnetp:
This is great.  But I have lots of questions.  Don't need to answer immediately as I know this is a work in progress.

* If this is merged are we committed to LTI in the sense that other grading functions will not work withour own database anymore?
* How does merging this effect login?
* It looks like this implementation assumes a 1-1 relationship between the runestone server and an LTI instance.  Thats a great start for people who run their own server.  But may not be the solution for our textbook as a service.
* Where / how do we now create assignments?  Do we continue to create them on runestone?  Do we create them in our LMS?  Can they be created in either place?
* Which LMS will work with this?  Canvas? Moodle? Others?

presnick:
This doesn't change anything about grading in runestone. It's just that when a student's assignment score is recorded, if the student ever previously accessed the assignment through the LTI interface, it sends the grade back to the LMS.

> How does merging this effect login?
Should not affect the ability to do normal logins to runestone. But it offers the option of logging in from the LMS. The LMS sends an email address and we now use that as the runestone username. This creates a small security hole compared to my old mechanism of generating a horrible username, because any LMS that you've trusted by giving it a set of LTI keys will be able to let a user login to the runestone server with any email address that's valid in the LMS. So we may want to revisit that issue again.

> It looks like this implementation assumes a 1-1 relationship between the runestone server and an LTI instance. Thats a great start for people who run their own server. But may not be the solution for our textbook as a service.

Sort of. You can have a separate LTI key and secret for each LMS instance that you want to give access to the runestone server. That's the lti_keys table.  But like with login, there isn't any security beyond that. Probably we should add a course_id field to the lti_keys table, and only allow login to the course_id associated with the key and secret.

> Where / how do we now create assignments? Do we continue to create them on runestone? Do we create them in our LMS? Can they be created in either place?

Right now, we have to create the assignment in both places, unfortunately. First, I create it in runestone, and note its assignment id. Then, in Canvas, I create an assignment and configure the external lti link, with url path `/runestone/lti?assignment_id=<id>`. There is an LTI feature, I think, for doing something on the runestone side that automatically creates an assignment object in the LMS, but I haven't implemented that yet.

> Which LMS will work with this? Canvas? Moodle? Others?

External login has been tested with Canvas, Sakai, and Blackboard. The gradeable assignment links have been tested only in Canvas. In principle, it should work with all the LTI-compliant LMSes.  To get started with using this on an LMS, an "external app" has to be configured. Eventually, we should implement a URL endpoint that will provide an XML configuration file back to the LMS. For right now, we've been doing that manually. A sample of what the XML configuration file looks like is in the file config.lti.

models/db.py
```diff
@@ -77,11 +78,12 @@
...
-    db.courses.insert(course_name='boguscourse', term_start_date=datetime.date(2000, 1, 1)) # should be id 1
-    db.courses.insert(course_name='thinkcspy', base_course = 'thinkcspy', term_start_date=datetime.date(2000, 1, 1))
-    db.courses.insert(course_name='pythonds', base_course = 'pythonds', term_start_date=datetime.date(2000, 1, 1))
-    db.courses.insert(course_name='overview', term_start_date=datetime.date(2000, 1, 1))
-    db.courses.insert(course_name='publicpy3', base_course='pip2', term_start_date=datetime.date(2000, 1, 1))
+    db.courses.insert(id=1, course_name='boguscourse', term_start_date=datetime.date(2000, 1, 1)) # should be id 1, to avoid error in auto populate of auth_users, where course_id defaults to 1
+    db.courses.insert(id=2, course_name='thinkcspy', base_course = 'thinkcspy', term_start_date=datetime.date(2000, 1, 1))
+    db.courses.insert(id=3, course_name='pythonds', base_course = 'pythonds', term_start_date=datetime.date(2000, 1, 1))
+    db.courses.insert(id=4, course_name='overview', term_start_date=datetime.date(2000, 1, 1))
+    db.courses.insert(id=5, course_name='publicpy3', base_course='pip2', term_start_date=datetime.date(2000, 1, 1))
```
bnmnetp: These probably don't belong in the general commit.   Also hardcoding the id's is going to get courses_id_seq out of sync.  If the database is empty then these courses will insert with the ids you want anyway.  So I can't imagine why you would need to hardcode them.

PR #1064, opened 2017-12-28
Unittest grading

presnick:
Really just one commit, with slight improvements to the shell script. But also has all my lti updates . If you don't want to accept those, please just cherry-pick c61014e

BTW, the quotes around $1 are needed in case there is no parameter provided on the command line, to make it still be an empty string.


bnmnetp:
Paul, I think you want to close this and do your unittest work on a new branch.  This seems to have a bunch of the LTI stuff mixed in.

presnick:
OK, will do the cherry-pick and resubmit.

PR #1072, opened 2018-01-02
control migration from environment variables; get better naming of th…

### LINKED ISSUE #1071: Control whether migrations run with environment variable
We are moving towards more control over migrations. Currently, in db.py we have:

```
    if os.environ.get("WEB2PY_CONFIG","") == 'test':
        db = DAL(settings.database_uri,migrate=False,migrate_enabled=False)
    else:
        db = DAL(settings.database_uri,fake_migrate_all=False)
```
Because our Docker-based deployment is not saving the databases/ folder between deployments, I would like to run migrations only when I explicitly ask for them.

How about this?

```
    if os.environ.get("WEB2PY_CONFIG","") == 'test':
        db = DAL(settings.database_uri,migrate=False,migrate_enabled=False)
    else:
        db = DAL(settings.database_uri, fake_migrate_all=os.environ.get("WEB2PY_FAKE_MIGRATE", False), migrate=os.environ.get("WEB2PY_MIGRATE", True), migrate_enabled=os.environ.get("WEB2PY_MIGRATE", True) )
```

presnick:
…e last few session_named metadata files

fixes #1071

### DIFF
--- a/models/db.py
+++ b/models/db.py
@@ -18,8 +18,10 @@ if not request.env.web2py_runtime_gae:
     if os.environ.get("WEB2PY_CONFIG","") == 'test':
         db = DAL(settings.database_uri,migrate=False,migrate_enabled=False)
     else:
-        db = DAL(settings.database_uri,fake_migrate_all=False)
-    session.connect(request, response, masterapp='runestone', db=db)
+        # WEB2PY_MIGRATE is either "Yes", "No", "Fake", or missing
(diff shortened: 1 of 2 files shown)


bnmnetp:
Can you update the README to document this new WEB2PY_MIGRATE environment variable so that people who want to set it know that it exists and how they should set it.  Thanks.

PR #1075, opened 2018-01-04
allow for / in subchapter names

presnick:
Subchapter names are taken from headers that people write in .rst files. So we should allow for / in them. This optional parameter to split just divides the string at the first /.

### DIFF
--- a/controllers/assignments.py
+++ b/controllers/assignments.py
@@ -1244,7 +1244,7 @@ def doAssignment():
 
     for r in readings:
         logger.debug("READING = %s",r.name)
-        chapterSections = r.name.split('/')
+        chapterSections = r.name.split('/', 1)
 
         labels = db((db.chapters.chapter_name == chapterSections[0]) & \
                     (db.chapters.course_id == auth.user.course_name) & \


bnmnetp:
If we are going to develop a culture of unit testing, this is a good example of where it would be ideal to construct a simple unit test that demonstrates the problem (ie fails on the current code) as well as the solution.

bjones1:
As I recall, it's not a big deal to me, but it's more of a concern for @presnick -- I can work with pretty much anything (label or name), but the back end needs to be able to map between labels and name. (Though I can't remember all the history here....I remember we changed from labels to names, but not why,...)

Re: unit testing, that's one thing that's been so hard in this project -- no tests to help me from introducing additional bugs in the code. IMHO one of the biggest contributions from the hack session was Jackie and (can't remember his name) Selenium tests. I've used those again and again and they've saved me from making many mistakes.

We need to include the server and the client (RunestoneComponents) in the tests, since they'll all involved in these sort of things. One of my goals for the NewServer / Flask port is unit tests for everything.

bjones1:
What unit tests are currently available? I'd certainly be happy to contribute something, but I wasn't aware of anything but the Selenium tests...

bnmnetp:
One of my New Years resolutions is to write one unit test a day between now and June.  Even if its a really simple one.  Its amazing the inconsistencies I've already uncovered.  So far I've just been adding #TODO style comments to things that need to be fixed in the tests themselves.  But I think getting  a decent baseline is critical.

bnmnetp:
@bjones1 the RunestoneServer now has a tests folder.   You can take a look at test_ajax.py in that folder to get an idea of how easy it is to write some simple tests.

bjones1:
Neat! Thanks. I'll take a look.

presnick:
I agree that a unit test for this would be great. But I can't take the time right now. The case that brought this to my attention was:

chapter filename `Exceptions/using-exceptions.rst` with header
```
When to use try/except
----------------------
```
That created this in the questions table:
```
si106w18=> select id, base_course, name, chapter, subchapter from questions where id=2635;
  id     | base_course |               name                                  |  chapter     |    subchapter
------+-------------+-----------------------------------+------------+------------------
 2635 | pip2              | Exceptions/When to use try/except | Exceptions | using-exceptions

si106w18=> select * from sub_chapters where id = 5297;
  id  | sub_chapter_name | chapter_id | sub_chapter_length | sub_chapter_label
------+------------------+------------+--------------------+-------------------
 5297 | Extra Exercises  |        711 |                    | ExtraExercises
```

bnmnetp:
I hear you.  Certainly adding two characters to one line of code takes a lot less time.  And I'm sure there is urgency in getting your course running before classes start.

I'm not throwing stones at you as I'm just as guilty (more so), but the challenge is how do we get to a place where "I don't have the time to write a test case" is not an acceptable answer?  We don't want to discourage PR's by having too many rules,  but as Brian points out, and as I've read more about successful open source communities, providing ways for people to feel comfortable that their contributions are breaking things is pretty important.

* There are definitely projects that would simply not accept the PR until there was a test case. -- Skulpt is like that as are many others.
* We could also make a rule that says if two out of three "committers" agree then we can accept a PR without.  -- That would probably apply here.  It also assumes that more people than just me have the power to accept a PR.

bnmnetp:
To Keep my New Years resolution I wrote a test case to demonstrate the problem here.  I've also incorporated Paul's fix to show that the fix works.  So, I'll close this without merging as the fix will be part of my commit.  For posterity, here is the test I wrote:

```python
    def testASlashInSubchapter(self):
        auth.login_user(db.auth_user(11))
        bad_name = 'Exceptions/When to use try/except'

        chap_id = db((db.chapters.chapter_name == 'Exceptions') & (db.chapters.course_id == auth.user.course_name)).select(db.chapters.id).first()
        db.sub_chapters.insert(sub_chapter_name='When to use try/except', chapter_id=chap_id,
                               sub_chapter_length=0, sub_chapter_label='using-exceptions')

        db.questions.insert(base_course='thinkcspy', name=bad_name,
                            chapter='Exceptions',
                            subchapter='using-exceptions',
                            htmlsrc='<h2>Hello World</h2>', question_type='page')

        request.vars.due = '2018/12/12 17:30'  # TODO: finish PR to use dateutil to parse dates in createAssignment
        request.vars.name = 'badsubchapter_assignment'
        res = createAssignment()  # TODO: deprecated?? but still used in admin.js

        res = json.loads(res)
        assignment_id = int(res[request.vars.name])

        request.vars.assignment = assignment_id
        request.vars.question = bad_name
        request.vars.points = 5
        request.vars.autograde = 'interact'
        request.vars.which_to_grade = ''
        request.vars.reading_assignment = True
        request.vars.activities_required = 2

        res = add__or_update_assignment_question()
        res = json.loads(res)
        rows = db(db.assignment_questions.assignment_id == assignment_id).select()
        self.assertEqual(1, len(rows))
        self.assertEqual(res['total'], rows[0].points)
        self.assertEqual('interact', rows[0].autograde)
        self.assertEqual(2, rows[0].activities_required)
        # select * from assignment_questions where assignment_id = 1244;
        #   id   | assignment_id | question_id | points | timed | autograde | which_to_grade | reading_assignment | sorting_priority | activities_required
        # -------+---------------+-------------+--------+-------+-----------+----------------+--------------------+------------------+---------------------
        #  30325 |          1244 |       14561 |      5 |       | interact  |                | T                  |                0 |                   2

        request.vars.assignment_id=str(assignment_id)
        res = doAssignment()
        # TODO: verify results of doAssignment
```

presnick:
I'm not sure what this is testing. I think the problem we were experiencing was from `Exceptions/When to use try/except` in the questions.name field. That isn't in the data you're inserting. Also, it assumes that the Exceptions chapter already is in the DB, which might not be the case.

bnmnetp:
Yes, I'm inserting that question into our test database near the top of the function.  I had to add this one since our test database does not have it.

PR #1076, opened 2018-01-07
Upstream updates

presnick:
Fixes bugs in rsmanage.py

### DIFF
--- a/rsmanage/makeuser.py
+++ b/rsmanage/makeuser.py
@@ -6,7 +6,7 @@ def createUser(username, password, fname, lname, email, course_name, instructor=
     cinfo = db(db.courses.course_name == course_name).select().first()
     if not cinfo:
         raise ValueError("Course {} does not exist".format(course_name))
-    pw = CRYPT(auth.settings.hmac_key)('grouplens')[0]
+    pw = CRYPT(auth.settings.hmac_key)(password)[0]
     uid = db.auth_user.insert(username=username, password=pw, first_name=fname,
                         last_name=lname, email=email,
                         course_id=cinfo.id, course_name=course_name, active='T',
--- a/rsmanage/rsmanager.py
+++ b/rsmanage/rsmanager.py
@@ -146,7 +146,9 @@ def addcourse(config, course_name, basecourse, start_date, python3, login_requir
         if not course_name:
             course_name = click.prompt("Course Name")
         if not python3:
-            python3 = 'T' if click.confirm("Use Python3 style syntax?", default=True) else 'F'
+            python3 = 'T' if click.confirm("Use Python3 style syntax?", default='T') else 'F'
+        else:
+            python3 = 'T'
         if not basecourse:
             basecourse = click.prompt("Base Course")
         if not start_date:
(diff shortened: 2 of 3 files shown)


rsmanage/makeuser.py
```diff
@@ -6,7 +6,7 @@ def createUser(username, password, fname, lname, email, course_name, instructor=
     cinfo = db(db.courses.course_name == course_name).select().first()
     if not cinfo:
         raise ValueError("Course {} does not exist".format(course_name))
-    pw = CRYPT(auth.settings.hmac_key)('grouplens')[0]
+    pw = CRYPT(auth.settings.hmac_key)(password)[0]
```
presnick: Was hardcoding `grouplens` as the password for everyone...

rsmanage/rsmanager.py
```diff
@@ -146,24 +146,28 @@ def addcourse(config, course_name, basecourse, start_date, python3, login_requir
         if not course_name:
             course_name = click.prompt("Course Name")
         if not python3:
-            python3 = 'T' if click.confirm("Use Python3 style syntax?", default=True) else 'F'
+            python3 = 'T' if click.confirm("Use Python3 style syntax?", default='T') else 'F'
+        else:
+            python3 = 'T'
```
presnick: Need 'T' or 'F' in DB insert, not True or missing. Concern: is 'T' and 'F' for booleans specifically a PSQL thing, or is that generic for all SQL?

bnmnetp: You don't want the else there -- the way this is meant to work is that python3 either be defined by a command line parameter, or the user will be prompted for it.

Postgresql conforms to the SQL standard in how it implements the boolean data type.

rsmanage/makeuser.py
bnmnetp: Oops, nice catch

rsmanage/rsmanager.py
```diff
@@ -146,24 +146,28 @@ def addcourse(config, course_name, basecourse, start_date, python3, login_requir
...
         if not basecourse:
             basecourse = click.prompt("Base Course")
         if not start_date:
             start_date = click.prompt("Start Date YYYY-MM-DD")
         if not institution:
             institution = click.prompt("Your institution")
         if not login_required:
-            login_required = 'T' if click.confirm("Require users to log in", default=True) else 'F'
+            login_required = 'T' if click.confirm("Require users to log in", default='T') else 'F'
+        else:
```
bnmnetp: Same comment as for python3

rsmanage/rsmanager.py
presnick: Which else do you think is unneeded? I think line 150-151 are needed because when python3 is set as a command flag, its value is True rather 'T'.

bnmnetp: Ahh I think you are right.... I

f the --python3 option is omitted then the python3 parameter will be None which causes click.confirm() to ask the user.  It can return true as the conditional assignment uses 'T' so you don't need to change the default  to 'T' but that probably doesn't matter.

But if the --python3 flag is set, it needs to be changed to 'T' rather than True otherwise the insert will fail.

I swear I testing this subcommand and it worked... but looking at the execute statement I can't see how.

I suppose I really should clean this up by using the declarative model of SQLAlchemy and having using the constructor to add the row.

Issue #1093, opened 2018-01-29
Incorrect parameter when grading


bjones1:
I think I've tracked down a subtle bug in assignment grading, but wanted to confirm. I'm getting an `IntegreityError` in [_save_question_grade](https://github.com/RunestoneInteractive/RunestoneServer/blob/master/controllers/assignments.py#L666). Tracking this down, the `db.question_grades.update_or_insert` call has `useinfo_id = useinfo_id`, and the ID being passed is invalid. Per the [question_grades table](https://github.com/RunestoneInteractive/RunestoneServer/blob/master/models/grouped_assignments.py#L545), this should be an entry in the `useinfo` table. However, the ID passed is actually an ID from whatever table held the relevant answer (in my case, `fitb_answers`): [here](https://github.com/RunestoneInteractive/RunestoneServer/blob/master/controllers/assignments.py#L662) is the call; at this point, `id` is set from a row of `results`; `results` comes from a `_scorable_xxxx` function, which returns a question-specific table, not a `useinfo` entry. For example, [here](https://github.com/RunestoneInteractive/RunestoneServer/blob/master/controllers/assignments.py#L607) is the call for a fill-in-the-blank question, which [queries](https://github.com/RunestoneInteractive/RunestoneServer/blob/master/controllers/assignments.py#L526) the `fitb_answers` table.

So, what's the best fix? Is this field (which has obviously been invalid for a bit) even used anywhere? Finding the correct `useinfo` id seems non-trivial to me, since there's no direct ID in each `xxx_answers` table back to `useinfo`. Searching the code, I can only find `useinfo_id` being used by `_save_question_grade`. For a temporary fix, I set `useinfo_id=None` in the update, and can now at least grade assignments without an exception.

bnmnetp:
I have no idea what the useinfo_id field is used for. I’ve wondered for a while but never dug into it.

bjones1:
Unless @presnick has a specific purpose, let's set it to `None` to avoid errors.

presnick:
I think there was some vague notion that we wanted to have a record of what run was graded. It is probably left over from when autograding only worked on activecodes and thus we didn't notice the problem. I agree that setting it to None is the best quick solution.

bjones1:
@bnmnetp, do you want a PR for this, or will you just make the edit yourself?

bnmnetp:
I would be happy to get a PR.

bjones1:
Closing this for the fix in #1094.

PR #1094, opened 2018-01-30
Fix: Avoid invalid useinfo_id.

bjones1:
See https://github.com/RunestoneInteractive/RunestoneServer/issues/1093.

### DIFF
--- a/controllers/assignments.py
+++ b/controllers/assignments.py
@@ -639,31 +639,27 @@ def _autograde_one_q(course_name, sid, question_name, points, question_type, dea
             else:
                 # default is last
                 row = results.last()
-            # extract its score and id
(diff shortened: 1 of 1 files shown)


bnmnetp:
Thanks @bjones1 -- Have you done grading on your installation with this PR?  We don't have a unit test that tests this path yet.  Want to write one 😁 ??

bjones1:
Yes, this works on my install; without this, I get an exception.

Sigh -- I'd like to write tests, but lack the time during the semester to do this.

bnmnetp:
I added a test

bjones1:
Wow, way to go! That's helpful for me to add tests in the future...

PR #1112, opened 2018-06-18
WIP: The ability to auto-route to a book based on URL and the courses table.

bjones1:
This is a WIP for an exciting idea: instead of creating an entire copy of a book every time a new course is added, simply look up the base course and refer to the base course files.

To set this up, there **must** be a `routes.py` file in the the root `web2py/` folder which contains:

``` Python
routes_app = (
    ('/runestone/.*', 'runestone'),
)
```

WIP means there's more to do:
- [x] The code at present treats all files as HTML, though it should return CSS, JS, etc. differently.
- [x] Performance. Will this be fast enough? Find nice ways to route static files?
- [ ] Nicer error checking and messages.
- [x] Templates. This should fill in the JS header of each page to specify the base course, but it doesn't. Ic particular, surround `_templates\plugin_layouts\sphinx_bootstrap\layout.html` code that sets up `eBookConfig` with `{% raw %}` and `{% endraw %}` tags to keep Sphinx from interpreting it, then update the variable names to match what's passed in the dict.
- [x] Redo the New Course Page
- [x] Remove the tab to Rebuild a course
- [x] Remove the tab to edit the TOC
- [x] Logging in is database driven according to the course.  Remove the @auth.requires_login() from `published` and add some logic to check whether a login is required for this book.
- [x] Update the navbar template so that we can insert the name of the course at runtime instead of the Book's Title

-------

Post Accept thing to do.

- [ ] All of the login stuff that is now done in javascript on the client side can be revamped.  Should be done in its own PR as it will touch a lot of places, and we need a backward compatibility period for courses in progress.


bnmnetp:
For static files, the fastest solution is going to be to configure nginx to respond to those requests directly without even bothering web2py.  If I recall correctly nginx has a decent pattern matching setup to allow you to specify those folders.

bjones1:
That seems reasonable. Some sort of script would need to regenerate this routing file when a new course is added, which shouldn't be too hard to do.

bjones1:
Some ideas for improving this:
- For static files, `return response.stream('/path/to/static/file', 2**20)`.
- For web pages, set `response.view = '/path/to/html/file'`, then return a dict to be rendered.

bjones1:
This works now! It's the start of a much more flexible approach. But, there's a lot more to decide on. @bnmnetp and @presnick, I'd like to have a conversation about this when yall have time.

bjones1:
The most important item in terms of moving in this direction is performance. If the servers can run this, it brings a lot of benefits. I think that allowing the web server to directly serve static files would be good enough -- even though serving an individual page will be longer, I'm guessing this won't be a bit factor in overall server load.

So, what options do we have to make it easy for a server to serve static files directly?

- A big regex. If there are hundreds or thousands of courses, the regex would have to contain hundreds or thousands of patterns to identify all the base course of many courses. This seems like a bad idea.
- Automatically editing the resulting HTML. It would probably be possible to patch up all references to `_static` in the `<head>` of each document during `runestone deploy` or something similar. Making the references be `../base_course_path/_static` would work, of course. However, I'm not excited about this approach.
- Naming the books to make the base course name easy to find. For example, if the href was `http://hostname/runestone/base_course/course/_static/file.js`, then a simple regex could serve this statically. I don't like repeating a database entry (the base_course for a given course_name), but it's not too bad. Other file naming ideas (a prefix to each course_name to refer to the base_course, for example.)
- Something else? Ideas are welcome.

bjones1:
This is now done. I'll push the changes after other PRs have been merged. Static routing works easily on Apache; the config I'm using is:

```
  AliasMatch ^/([^/]+)/books/draft/([^/]+)/_static/(.*) \
        /home/www-data/web2py/applications/$1/books/$2/build/$2/_static/$3
  <Directory /home/www-data/web2py/applications/*/books/*/build/*/_static/>
    Options -Indexes
    ExpiresActive On
    ExpiresDefault "access plus 1 hour"
    Require all granted
  </Directory>

  AliasMatch ^/([^/]+)/books/published/([^/]+)/_static/?(.*) \
        /home/www-data/web2py/applications/$1/books/$2/published/$2/_static/$3
  <Directory /home/www-data/web2py/applications/*/books/*/published/*/_static/>
    Options -Indexes
    ExpiresActive On
    ExpiresDefault "access plus 1 hour"
    Require all granted
  </Directory>
```

bjones1:
This is blocked by #1216.

bnmnetp:
OK, so this is unblocked now, but I suspect you have more parts to this that you haven't pushed to update this PR??    I'm able and happy to work on this now.

bjones1:
I just pushed what I have, which is mostly complete. Some of the routing for the static books needs fixing, but it's fairly close. (This is what I use on my production server.)

I should be able to update this tomorrow.

bnmnetp:
Great, thanks.  This and 3.0 are my top priorities, as I think  this will change how many new features could be implemented.

bjones1:
OK, this is pretty close. Things that still need doing:

- [ ] Fix the test failure -- looks like the test env isn't setting `request.folder` or `auth.user.course_name`.
- [ ] Add a book that's statically-routed to the tests, then hit its endpoints.

models/db.py
```diff
@@ -89,6 +89,26 @@
...
+    if not args:
+        args = db.courses.course_name, db.courses.base_course
+    return db(db.courses.id == auth.user.course_id).select(*args).first()
+
+
+# Provide the correct URL to a book, based on if it's statically or dynamically served. This function return URL(*args) and provides the correct controller/function based on the type of the current course (static vs dynamic).
+def get_course_url(*args):
+    # Redirect to old-style statically-served books if it exists; otherwise, use the dynamically-served controller.
+    if os.path.exists(os.path.join(request.folder, 'static', auth.user.course_name)):
+        return URL(c='static', args=[auth.user.course_name, *args])
```
bnmnetp: I'm getting syntax errors on this -- `args=[auth.user.course_name] + list(args)` does what I **think** you want...

bjones1: That sounds reasonable. I didn't test under Python 2, perhaps that's why I missed it.

bjones1: Yep, confirmed in CI -- 2.7 fails here, 3.7 passes it.

models/db.py
```diff
@@ -89,6 +89,26 @@
...
+def get_course_row(*args, **kwargs):
+    if not args:
+        args = db.courses.course_name, db.courses.base_course
+    return db(db.courses.id == auth.user.course_id).select(*args).first()
+
+
+# Provide the correct URL to a book, based on if it's statically or dynamically served. This function return URL(*args) and provides the correct controller/function based on the type of the current course (static vs dynamic).
+def get_course_url(*args):
+    # Redirect to old-style statically-served books if it exists; otherwise, use the dynamically-served controller.
+    if os.path.exists(os.path.join(request.folder, 'static', auth.user.course_name)):
```
bnmnetp: In the older style of tests, you have to explicitly setup all the properties of requests.... We had no reason to setup  requests.folder prior to this

bjones1: That what I expected. Another reason to port these tests as well...

Can you work on this? I'm busy until next Tuesday with grading/turning in final grades.

bnmnetp: Yes, I fixed that and the syntax error and tests are now passing except for a mysterious

```
Ran 7 tests in 9.051s

OK
  File "/srv/web2py/gluon/shell.py", line 282, in run
    execfile(startfile, _env)
  File "applications/runestone/tests/test_assignments.py", line 264, in <module>
    sys.exit(0)
SystemExit: 0
```
At the end of each test.   I don't remember that showing up before, so I'm not sure why it started now.

bjones1: Great, thanks! Fantastic.

I agree -- this is odd. I don't remember seeing that before either.

bnmnetp: I wish there was some way I could update your PR

bjones1: Is there anything I can do on my side? Give you permission to my repo, I suppose?

bnmnetp:
@bjones1 — lots of good stuff in here...  I’m getting hundreds of 404’s on the instructor interface because it is looking for stuff in /books/...  Should I be rebuilding the base courses some way ? Or is there some command I should be running to get these in place - The performance of the instructor interface is also super slow which I’m putting down to the missing files...  

Also I’m noticing that many of the links that should take one back to the book are incorrect. There’s an extra copy of the current controller always inserted before /static.

I’m reminded that there is a lot of cleanup to do in instructors.html template duplicate information, extra information.... junk like that that has accumulated over the years.

bjones1:
Thanks for taking a look. Yes, all the dynamic book stuff requires the base course to be built for dynamic serving, based on https://github.com/RunestoneInteractive/RunestoneComponents/pull/805. You'll need to re-create the book using `runestone create`, then copy the book files in since that will updated `pavement.py`, `config.py`, and lots of the Sphinx templates. Then, everything should work. Note that `test_10` in `test_server.py` tests this, so it should be working.

The slowness comes from that fact that **ALL** pages -- including all the static JS and images -- go through web2py as a full request, not a static request. Setting your web server up to server the static pages makes this very snappy again (since the only thing web2py servers is the actual HTML of a given page, not all the JS and images).

bjones1:
Aargh, I see a horrible mistake. Line 113 of `db.py` should be:

``` Python
            return URL(c='books', f='published', args=[course.base_course] + list(args))
```

I'm not sure how to submit this as a PR (open a second PR on your `bjones1-book-routing` branch?) I **do** have the "allow edits from maintainers" box checked, so you should be able to edit my PRs...

bnmnetp:
Can’t you just add this to your existing PR?  I don’t think that branch is relevant as I was able to push to your Fork with my updates...

bjones1:
Good point -- done. Let me know if you see anything else broken.

bnmnetp:
I'm not seeing a new command `runestone create` in that PR.  It looks to me like what I really need to do is update the layout.html template and then rebuild and redeploy the book to the right place.

views/_sphinx_static_files.html
```diff
@@ -1,87 +1,88 @@
...
+<script src="{{=base_static_url}}/activecode.js" type="text/javascript"></script>
+<script src="{{=base_static_url}}/mchoice.js" type="text/javascript"></script>
+<script src="{{=base_static_url}}/shortanswer.js" type="text/javascript"></script>
+<script src="{{=base_static_url}}/js_lib/prettify.js" type="text/javascript"></script>
+<script src="{{=base_static_url}}/js_lib/hammer.min.js" type="text/javascript"></script>
+<script src="{{=base_static_url}}/parsons.js" type="text/javascript"></script>
+<script src="{{=base_static_url}}/fitb.js" type="text/javascript"></script>
+<script src="{{=base_static_url}}/dragndrop.js" type="text/javascript"></script>
+<script src="{{=base_static_url}}/timeddnd.js" type="text/javascript"></script>
+<script src="{{=base_static_url}}/lp.js" type="text/javascript"></script>
```
bnmnetp: This `lp.js` file appears to be missing from the PR or else it might be a typo??

bjones1: Oops, thanks. That belongs in my next PR (adding literate programming). I'll delete it for now.

bnmnetp:
So I have books serving nice and fast from this PR now with nginx.  In the site file for runestone I just have this

```
    # alias allows us to completely rewrite the static path
    location ~* /(\w+)/books/published/(\w+)/_static/(.*)$ {
        alias /srv/web2py/applications/$1/books/$2/published/$2/_static/$3;
    }
```

However... 

1. When I register as a new user I'm routed to /runestone/static/fopp/....  I guess because it exists in both places?

2.  When I try to create a new course based on fopp I get a crash

```
---------------- Traceback --------------
  File "/srv/web2py/gluon/restricted.py", line 219, in restricted
    exec(ccode, environment)
  File "/srv/web2py/applications/runestone/controllers/designer.py", line 183, in <module>
  File "/srv/web2py/gluon/globals.py", line 422, in <lambda>
    self._caller = lambda f: f()
  File "/srv/web2py/applications/runestone/controllers/designer.py", line 55, in build
    uuid = row['uuid']
UnboundLocalError: local variable 'row' referenced before assignment
```
Is there work to be done here or should this just work??

bnmnetp:
That appears to be a line that should have been deleted when the query to the scheudler was called.

bnmnetp:
@bjones1 - How is adding a new course based on a basecourse supposed to work?  I assumed I would just fill out the form and a database entry would be made, but it sends me to the 'building a course' page needlessly.

bjones1:
@bnmnetp, sorry, I've been out of the office most of the morning. Some quick answers:

1. Until you rebuild the base course using the new template and remove it from `web2py/applications/runestone/static/fopp` (or whatever the course name is), the system will prefer a staticly-built class to a dynamic class.
2. Oops -- thanks for finding my bug. I didn't even think to test this code --I've always just added an entry to the `courses` table manually, which is all it takes to create a new dynamic course. So, I suspect this part is now broken -- the course is available immediately after the database is updated, so there's no need for a building page.

bjones1:
@bnmnetp, for your regexp, a few notes:

1. You need another regexp for `books/draft` (an instructor-only view of the `build/` directory produced by `runestone build`.
2. You need another two regexps for `_images`, another path Sphinx uses to store static content. (It might be `_image`, I'm away from my desk where I could check this -- just take a look at the output from a `runestone build`.

bnmnetp:
OK, I can take care of the designer.py stuff.

bnmnetp:
So, I created a little `locustio` script to do some load testing... I'll commit it to the repo soon.

I noticed that there were a few failures in loading ordinary html pages... And after some work tracked it down to

```
<script>
    $(document).ready(function() {
      steps = ['x = {{x}}{{6}} + 1', 'x = {{6 + 1}}{{7}}'];
      se_ac2_14_1_object = new SHOWEVAL.ShowEval($('#se_ac2_14_1'), steps, true);
      se_ac2_14_1_object.setNextButton('#se_ac2_14_1_nextStep');
      se_ac2_14_1_object.setResetButton('#se_ac2_14_1_reset');
    });
</script>
```

Ugh, the web2py template is trying to expand `{{x}}`

bnmnetp:
Maybe not quite as bad as I thought at first... It seems that escaping the `{` as `\{` is a workaround.  We could do the escaping in the showeval directive. 

```
'x = \{\{x}}\{\{6}} + 1', 'x = \{\{6 + 1}}\{\{7}}'
```

bjones1:
Wow, I'm excited about the locust script for load testing. I meant to do that, but got busy. I'm glad you did.

I didn't think about code that's now being interpreted by web2py. Thanks for all your testing.

How about:

```
{{=r"""Anything here won't be interpreted by web2py! 
steps = ['x = {{x}}{{6}} + 1', 'x = {{6 + 1}}{{7}}'];
"""}}

I'm out of the office tomorrow and all next week.

bnmnetp:
OK, I updated the todo list at the top of this conversation.  I'll keep testing and working on things and update your fork as I go.

bnmnetp:
OK, this is getting a little unwieldy and it seems to work well now.  I've run through a bunch of workflows, you are/were using in production. 

The problems that I have run into recently seem to be more related to Python 3 than dynamic routing.  We can still hash out and make changes over the next weeks, but I think that will be easier to do as smaller PR's rather than continually adding to this one.

Time to merge.

PR #1116, opened 2018-06-22
Misc fixes

bjones1:
Another grab bag of fixes, with one minor new feature (an added column, the student's e-mail address, for the gradebook).


models/questions.py
```diff
@@ -12,7 +12,6 @@
                 Field('htmlsrc', type='text'),
                 Field('practice', type='boolean'),
                 Field('autograde', type='string'),
-                Field('practice', type='boolean'),
```
bnmnetp: this field is needed for the new spaced practice feature

controllers/ajax.py
```diff
@@ -77,6 +80,10 @@ def hsblog():
             logger.debug('correct {} incorrect {} skipped {} time {}'.format(request.vars.correct, request.vars.incorrect, request.vars.skipped, request.vars.time))
             logger.debug('Error: {}'.format(e.message))
 
+    # Produce a default result.
```
bnmnetp: Something here is breaking the unit tests.  From the tests folder can you run the run_tests script?  You can see the breakage on github where I have travis set up to automatically run the unit tests on all PR's

controllers/ajax.py
```diff
@@ -27,9 +27,10 @@
                'clickablearea':'clickablearea_answers',
                'parsonsprob': 'parsons_answers' }
 
+
 def compareAndUpdateCookieData(sid):
     if request.cookies.has_key('ipuser') and request.cookies['ipuser'].value != sid:
-        db.useinfo(db.useinfo.sid == request.cookies['ipuser'].value).update(sid=sid)
+        db.useinfo.update_or_insert(db.useinfo.sid == request.cookies['ipuser'].value, sid=sid)
```
bnmnetp: Do we really want update or insert here??  I think update should be sufficient.

controllers/ajax.py
```diff
@@ -50,13 +51,15 @@ def hsblog():
     div_id = request.vars.div_id
     event = request.vars.event
     course = request.vars.course
-    ts = datetime.datetime.now()
+    # Get the current time, rounded to the nearest second -- this is how time time will be stored in the database.
+    ts = datetime.datetime.utcnow()
```
bnmnetp: all controllers should use utcnow instead of now.  I seem to have misplaced a PR I had prepared that updates all the controllers so we are consistent.

controllers/ajax.py
```diff
@@ -140,7 +147,6 @@ def hsblog():
             sid=sid, answer=act, div_id=div_id, timestamp=ts, course_name=course)
 
     response.headers['content-type'] = 'application/json'
-    res = {'log':True}
```
bnmnetp: It may just be that the test itself needs to be updated to reflect your addition of the timestamp to the default??

models/questions.py
bjones1: But it is needed twice? (It also appears in line 13 above).

controllers/ajax.py
bjones1: Yes, I see the breakage in Travis. I'm porting the test shell script to a Python script, so I can run it on my machine. However, I'm at a conference this week (I presented two papers based around Runestone!), so it will probably be next week.

controllers/ajax.py
bjones1: In the case that this is the first time this `sid` is encountered, then this must be an update. If the `sid` is already in the database, an update works fine. So, I have seen failures when the first case is encountered.

controllers/ajax.py
bjones1: Yes, that's all. I'll work on this when I can, probably next week.

models/questions.py
bnmnetp: No, we don't need it twice!  I didn't see that it was duplicated.  Amazing that it doesn't cause a database error.

bjones1:
OK, this should be ready to go.

One issue: the existing SQL code that runs reports some [errors](https://travis-ci.org/RunestoneInteractive/RunestoneServer/builds/401604691#L1992) -- these need fixing (but not by me...)

controllers/ajax.py
bnmnetp: Why did you need to add timestamp to the return value??

bjones1: Thanks for taking a look!

Client-side [code](https://github.com/RunestoneInteractive/RunestoneComponents/blob/master/runestone/common/js/runestonebase.js#L115) compares the server's date with the client's date to determine which is newer. If the client's clock isn't set correctly, this comparison will return an incorrect value. So, I'm having the server provide the time, then using this in client code to record the time, rather than the client's clock.

tests/run_tests.py
```diff
@@ -0,0 +1,69 @@
...
+    #       raise ValueError("unknown value for WEB2PY_CONFIG")
+
+    # HINT: make sure that you export TEST_DBURL in your environment; not set here because it's specific to local
+    # setup, possibly with a password, and thus can't be committed to the repo.
+    assert os.environ['TEST_DBURL']
+
+    # Extract the components of the DBURL. The expected format is ``postgresql://user:password@netloc/dbname``, a simplified form of the `connection URI <https://www.postgresql.org/docs/9.6/static/libpq-connect.html#LIBPQ-CONNSTRING>`_.
+    empty1, postgres_ql, pguser, pgpassword, pgnetloc, dbname, empty2 = re.split('^postgres(ql)?://(.*):(.*)@(.*)/(.*)$', os.environ['TEST_DBURL'])
+    assert (not empty1) and (not empty2)
+    os.environ['PGPASSWORD'] = pgpassword
```
bnmnetp: I don't understand why you don't just get both the password and the user from the TEST_DBURL?  Why make the user set more environment variables when everything we need should be there??

views/dashboard/index.html
```diff
@@ -22,7 +22,7 @@ <h2 style="width: 800px; display: inline;">{{=selected_chapter['chapter_name']}}
 					  {{pass}}
 					  {{pass}}
 					</select>
-					<button onclick="window.location.href='grades'">Gradebook</button>
+					<button onclick="window.location.href+='/grades'">Gradebook</button>
```
bnmnetp: @bjones I think this is a typo `+=` breaks the gradebook link.  After multiple clicks I end up with a URL like:  ``http://127.0.0.1:8000/runestone/dashboard/index/grades/grades/grades/grades/grades``

tests/run_tests.py
bjones1: This statement sets an environment variable that (on Windows) the Postgres command-line tools require to work. However, it's set based on the parsed value of `TEST_DBURL` -- the user doesn't have to set any additional command-line variables.

Or am I completely missing something?

bnmnetp: but the user is unconditionally set to 'postgres'

bnmnetp: I'm going to change it so that os.environ['PGUSER'] is set to pguser from the parsed url.

views/dashboard/index.html
bnmnetp: This change just needs to be reverted back to what it was.  Was it actually broken for you??  The old way was fine on runestone.academy and on my test machine.

bnmnetp: I'm reverting it in the test branch and will accept the pr with that change.

tests/run_tests.py
bjones1: This may be a Windows-only difference. Sorry, I just gave my first test (using Runestone!) so I'm been spending all my time on getting that working. I'll hopefully have time to think about this later today...

bnmnetp: Or maybe the user you set up in your URL does not have database superuser privileges??

views/dashboard/index.html
bjones1: It now works for me -- earlier, the link did't work at all. ??? Anyway, thanks!

PR #1118, opened 2018-07-04
some exception handling in case a question's topic does not match a c…

presnick:
…hapter and subchapter in db

### DIFF
--- a/controllers/admin.py
+++ b/controllers/admin.py
@@ -1099,17 +1099,26 @@ def _get_toc_and_questions():
                          (db.questions.practice == True) & \
                          (db.questions.topic != None)).select(db.questions.topic, orderby=db.questions.id)
         for q in topic_query:
-            chap, subch = q.topic.split('/')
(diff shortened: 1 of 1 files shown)


controllers/admin.py
```diff
@@ -1099,17 +1099,26 @@ def _get_toc_and_questions():
...
             # We have saved chapter_name and sub_chapter_name in db.sub_chapter_taught, and we know these names include
             # spaces in them. So we cannot directly use the labels retrieved from q.topic as chapter_name and
             # sub_chapter_name. So we need to query the corresponding chapter_name and sub_chapter_name from the
             # corresponding tables.
-            chapter = db((db.chapters.course_id == auth.user.course_name) & \
+            try:
+                chap, subch = q.topic.split('/')
+            except:
+                # a badly formed "topic" for the question; just ignore it
+                continue
```
bnmnetp: Should we at least log this error, or maybe even create a session.flash to tell the instructor that they should report a problem?  -- Shouldn't we really be detecting bad topics ahead of time?

PR #1123, opened 2018-07-12
Gdpr updates

bnmnetp:
This PR adds a database field to document registered users have accepted our privacy policy and enforces that the accept checkbox is clicked before allowing a user to register.

Further it adds a one-time-per-browser popup on our use of cookies and privacy policy.


bjones1:
I'd say leave it in regardless. It's getting to be an expected click, anyway. Also, what other policies do we want our users to agree too? If we go commercial, there's probably a license agreement. I have some legalese I bought from (I think) from rocketlawyer for all this I could contribute.

bnmnetp:
I have a privacy policy  (free template) as well as a terms and conditions document (probably also from rocketlawyer) in place on runestone.academy.

If you don't see any obvious problems with the PR I'll go ahead and merge this and get it live.

PR #1124, opened 2018-07-12
Fix: Get make_clean_db_with_grades.py working again.

bjones1:
I noticed that `run_test.py --rebuildgrades` failed; this fixes it.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,7 +23,10 @@ script:
 - git clone https://github.com/RunestoneInteractive/thinkcspy.git
 - cd ..
 - cd tests
+# Now actually run the tests.
 - python run_tests.py
+# Rebuild the grades just to make sure this works.
+- python run_tests.py --rebuildgrades
 notifications:
   slack:
     secure: OfeZ/+YHpSl7GNFkzEKu0uObuMdymKmDEeDyfB++RaglalcgkDQAqrCPx+YVw6brXOHGeC6BkGJe/Fs143Ai8/yf38nMmZg/dej5nLgZNST67jWaapIydx5QeuCrsRKvUP+z6yEYoBkYXTsYPyTiKGGsyzvX6dimZRT4nD8qrv0=
--- a/modules/rs_grading.py
+++ b/modules/rs_grading.py
@@ -1,11 +1,20 @@
 import datetime
 import logging
 from math import ceil
+
 from outcome_request import OutcomeRequest
+
+# When testing, the ``settings`` object isn't defined. Import it in this case.
+try:
+    logger
+except:
+    execfile('applications/runestone/models/0.py', globals())
+logger = logging.getLogger(settings.logger)
+logger.setLevel(settings.log_level)
 
 def _profile(start, msg):
     delta = datetime.datetime.now() - start
-    print "{}: {}.{}".format(msg, delta.seconds, delta.microseconds)
+    print("{}: {}.{}".format(msg, delta.seconds, delta.microseconds))
 
 
 def _score_from_pct_correct(pct_correct, points, autograde):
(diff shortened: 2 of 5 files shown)


.travis.yml
```diff
@@ -23,6 +23,9 @@ script:
 - git clone https://github.com/RunestoneInteractive/thinkcspy.git
 - cd ..
 - cd tests
+# Rebuild the grades just to make sure this works.
+- python run_tests.py --rebuildgrades
```
bnmnetp: We really shouldn't run rebuildgrades UNLESS we know we have intentionally changed something that changes how autograding works.  Otherwise we may be just hiding unintentional breakage by rebuilding.  If we do run rebuild grades then we should re-dump the database and update runestone_test.sql

modules/rs_grading.py
```diff
@@ -433,28 +442,24 @@ def send_lti_grade(assignment, student, lti_record, db):
         # send it back to the LMS
         # have to send a percentage of the max score, rather than total points
         pct = grade.score / float(points) if points else 0.0
-        # print "score", score, points, pct
+        # print("score", score, points, pct)
```
bnmnetp: Your adding the parenthesis makes me wonder if you are running web2py and the unit tests with python 3???

tests/make_clean_db_with_grades.py
```diff
@@ -12,7 +12,7 @@
 from gluon.globals import Request
 
 # bring in the assignments controllers
-execfile("applications/runestone/controllers/assignments.py", globals())
+execfile("applications/runestone/modules/rs_grading.py", globals())
```
bnmnetp: Ahhh, so when paul moved a bunch of stuff out of assignments.py and into the rs_grading.py module that was probably the source of the breakage.

bjones1: Yep -- that's the root of the problem.

modules/rs_grading.py
bjones1: I was confused by the same thing, and I can't really explain this. I printed `sys.version` just before one of these print statements and it's Python 2.7.13. I can run `python rs_grading.py` with no errors, but perhaps web2py's import machinery does something strange?

.travis.yml
bjones1: That makes sense. I still want to run the `--rebuildgrades` options just to make sure it runs, even if the output isn't used. Should I move this to be after the tests are run, then?

bnmnetp: Its fine to run it either before or after but just don't dump the database.

modules/rs_grading.py
bnmnetp: I wonder if there isn't a from __future__ import print_function somewhere in the mix then.  That is a good step in the direction of 3.

bjones1: That's what I'd expect.

.travis.yml
bjones1: I assume you mean that the code should not run [pg_dump](https://github.com/RunestoneInteractive/RunestoneServer/blob/master/tests/run_tests.py#L52)? If so, how do you want me to do this?

1. Add another command-line option that just run `make_clean_db_with_grades.py` but not `pg_dump`?
2. Do a `git checkout -- tests/runestone_test.sql` to revert any changes.
3. Run this after `python run_test.py`, so the changes don't matter.

The third choice seems simplest to me.

bjones1:
One motivation for me to do this was to support testing of another PR that will land after this one is merged, which requires updates to `runestone_test.sql`. What book was this database built against? I need to re-run a `runestone build` on it, since this incoming PR changes fields in the database and test code fails without those updates.

bjones1:
Fixed now!

bjones1:
@bnmnetp, pinging on this -- I think it's all ready to go...

bjones1:
Thanks!

PR #1128, opened 2018-07-28
Server-side feedback

bjones1:
This PR provides server-side feedback for fill-in-the-blank questions: rather than trusting the student's browser to grade a problem, the server performs the grading. 

Performance with this change will, of course, be an important question. I'd suggest providing a way to enable or disable server-side grading by adding an entry to the courses table (perhaps a `server_grading` boolean?). For now, server-side grading is enabled when `courses.login_required` is true.

To provide testing, this also include improvements to the tests; instead of relying on data in a test database, many tests now create the needed database entries then test using them. In addition, I've added a test book, making it easy to write test problems then verify their operation in the unit tests.

I'll be happy to submit a PR to RunestoneComponents with the client-side modifications needed to make this work when you're ready.


bjones1:
Oops. Of course, this can't work until the RunestoneComponents PR is merged. I'll put that together when I can.

bnmnetp:
@bjones1 several conflicts with files I'm surprised you needed to change.  I can't accept the PR on the components side until I have this in place.

bnmnetp:
Brian,

I'll investigate further.  I got panicked because I tried to build a book with the components pr merged and I was getting all kinds of database errors.  I can't take a chance on bringing down all of the assignments on runestone.academy so I assumed that these needed to come as a pair.

bjones1:
It's a chicken-and-egg problem, probably. The database needs one more field added to it, which means server. Then the components needs to run to populate that fiend. Then the server tests can run.

I'll also do some more investigating on my side and update the PR to resolve conflicts.

bjones1:
I **think** the only database field added was `Field('feedback', type='text'),` to `models\questions.py`. That add might make the components happy.

bjones1:
Closing this in favor of small PRs.

PR #1137, opened 2018-09-01
tweaks for install

dfrankow:
Tried installing. These things helped.
Also I had to make sure pip was new enough, otherwise it errored out.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -147,8 +147,8 @@ Installation
 
    * Tell web2py to use that database:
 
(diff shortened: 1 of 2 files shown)


dfrankow:
Ah, I see travis also makes private. Hmm. I guess I will remove that directory in git.
However, something I did wanted private/ and it didn't exist. I forget what. Oh well.

bnmnetp:
was it the ``rsmanage initdb`` command  maybe?

PR #1144, opened 2018-10-05
Misc updates

bjones1:
- Avoid calling `db.commit()` during an Ajax callback. This makes testing difficult (the database is modified unnecessarily.)
- Remove some `print` statements that produce noise in testing.
- Call `print` as a function, to prepare for porting to Python 3.

### DIFF
--- a/controllers/ajax.py
+++ b/controllers/ajax.py
@@ -408,13 +408,11 @@ def updatelastpage():
                    last_page_subchapter=lastPageSubchapter,
                    last_page_scroll_location=lastPageScrollLocation,
                    last_page_accessed_on=datetime.datetime.utcnow())
-        db.commit()
         db((db.user_sub_chapter_progress.user_id == auth.user.id) &
            (db.user_sub_chapter_progress.chapter_id == lastPageChapter) &
            (db.user_sub_chapter_progress.sub_chapter_id == lastPageSubchapter)).update(
                    status=completionFlag,
                    end_date=datetime.datetime.utcnow())
-        db.commit()
 
         practice_settings = db(db.course_practice.course_name == auth.user.course_name)
         if (practice_settings.count() != 0 and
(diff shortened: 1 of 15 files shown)


controllers/ajax.py
```diff
@@ -408,13 +408,11 @@ def updatelastpage():
                    last_page_subchapter=lastPageSubchapter,
                    last_page_scroll_location=lastPageScrollLocation,
                    last_page_accessed_on=datetime.datetime.utcnow())
-        db.commit()
         db((db.user_sub_chapter_progress.user_id == auth.user.id) &
            (db.user_sub_chapter_progress.chapter_id == lastPageChapter) &
            (db.user_sub_chapter_progress.sub_chapter_id == lastPageSubchapter)).update(
                    status=completionFlag,
                    end_date=datetime.datetime.utcnow())
-        db.commit()
```
bnmnetp: These were added to try to reduce the number of deadlocks that were occurring in the database under load.  It seems to me that they had the desired effect.   How does this mess up testing?  We could take them back out and see if the deadlocks come back.

bjones1: When testing, the best approach IMHO is to start with a clean database in a known state. My new testing addition (in a PR to come) call and Ajax function, then do a `db.rollback()` to discard any changes and keep the database in this known state.

So, this could be worked around in testing, if it make the production environment better, would would be a bit ugly. It's your call. If you want to omit these changes, I'll update the PR.

Thanks for the rapid feedback. That helps a lot.

bnmnetp: Lets go ahead and take them out.

bnmnetp: If I start to see crashes there again, then we will figure out a new solution

bjones1: Sounds good. So, as I understand it, this PR needs no changes?

bjones1:
Thanks!

PR #1146, opened 2018-10-08
Synthetic test database

bjones1:
This PR:
- Removed some unused imports in a few source files
- Adds a test book, to correctly populate the test database.
- Revised the tests to use synthetic test data where possible. This makes the test a lot easier to follow and update.
- Moves test-only dependencies to a separate `requirements-test.txt` file.

Other than the removed imports, this should have no impact on the production server -- this PR is focused on improving tests.


requirements-test.txt
```diff
@@ -0,0 +1,2 @@
+coverage
```
bnmnetp: @bjones1 This is great.  Maybe you could add a little bit to the run_tests.py program to try to import all the packages in requirements-test.txt . And stop if there is something missing. 

I had the old shell script set up so that if any of the individual test runs failed it would immediately stop so you knew there was an error.  So on my initial run of this new package I thought everything ran without problem but noticed that our coverage had dropped to 18%.  Scrolling back was the only way I discovered that test_ajax had failed becuause it was not able to import mock.

Does `xqt` return a status?  If so we should check it and bail if something fails.

bjones1: I did some investigating. When the tests complete, they call `sys.exit` with the appropriate return code. However, if the tests can't be imported by web2py, then web2py prints an exception but doesn't produce a non-zero exit status -- see https://github.com/web2py/web2py/blob/master/gluon/shell.py#L276.

I'm not sure how to detect this or work around it, however. Do you have any ideas? This is a **big** hole in testing. As far as I can tell, it exists independent of using `run_tests.py` or a shell script.

bjones1: OK, ugly hack: we could surround each test file with a try/except and call sys.exit. I'll try that out. Yuck.

bjones1: That works. But it's so ugly, there's got to be some better way. I'd suggest postponing a fix to a lter PR, since nothing in PR causes this bug, though it does reveal it.

bjones1:
Thanks!

PR #1147, opened 2018-10-09
Following Professor Resnick's advice, I added a condition to the  sta…

ImanYZ:
…tement to make sure it only sends LTI grades for those students who have a grade record for the assignment.

### DIFF
--- a/controllers/assignments.py
+++ b/controllers/assignments.py
@@ -632,8 +632,9 @@ def chooseAssignment():
 def _get_lti_record(oauth_consumer_key):
     return db(db.lti_keys.consumer == oauth_consumer_key).select().first()
 
-def _get_course_practice_record(course_name):
-    return db(db.course_practice.course_name == course_name).select().first()
+def _get_course_practice_record(course_name, user_id):
+    return db((db.course_practice.course_name == course_name) &
+              (db.course_practice.auth_user_id == user_id)).select().first()
 
 def _get_student_practice_grade(sid, course_name):
     return db((db.practice_grades.auth_user==sid) &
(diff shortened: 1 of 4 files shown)


controllers/assignments.py
```diff
@@ -693,6 +694,11 @@ def _get_qualified_questions(base_course, chapter_label, sub_chapter_label):
               (db.questions.practice == True)).select()
 
 
+# Gets invoked from lti to set timezone and then redirect to practice()
+def settz_then_practice():
+    return dict(course_name=request.vars.get('course_name', 'UMSI106'))
```
bnmnetp: You cannot hard code a course name (UMSI106) in code for general use.

controllers/assignments.py
```diff
@@ -632,8 +632,9 @@ def chooseAssignment():
 def _get_lti_record(oauth_consumer_key):
     return db(db.lti_keys.consumer == oauth_consumer_key).select().first()
 
-def _get_course_practice_record(course_name):
-    return db(db.course_practice.course_name == course_name).select().first()
+def _get_course_practice_record(course_name, user_id):
```
bnmnetp: Do we know that both course_name and user_id will be not None ?  What if they are?

controllers/assignments.py
ImanYZ: Professor @presnick has written this and he's not hard-coding it. He means if the 'course_name' does not exist, use 'UMSI106' as a default value. What else would you like to use as a default value?

controllers/assignments.py
ImanYZ: @bnmnetp, I do not call this function if either of them is none. You can verify this by looking at all the places that I've invoked this function.

controllers/assignments.py
bnmnetp: Well UMSI106 is only a Michigan course, so while it makes sense for a Michigan installation it really does not make sense for a general installation on runestone.academy

We really should have a setting in models/1.py  to provide a default value for this.

PR #1149, opened 2018-10-11
Html validation

bjones1:
This PR contains:

- Another big improvement in testing: the ability to run the web2py server, then send it requests and check the results programmatically. This will enable lots of good future testing of the server that we currently lack.
- HTML fixes. Much of the HTML is in sad shape: poorly indented, and with obvious errors that validators flag. This is the first step in HTML cleanup.

Note that this required another test dependency. Unlike the current system, if it's not installed, the tests will fail, since all testing takes place using a stand-alone unittest, instead of being run from inside web2py.


views/default/user.html
```diff
@@ -16,375 +16,342 @@
 
 <div class="container rscontainer">
 
-{{if request.args(0) != 'register':}}
```
bnmnetp: Is this just all stuff that was reformatted by some html tidy script or are there real and substantial changes in here?  This is the "front door" to runestone so huge changes like this with no context make me nervous?

bjones1: I did a lot of manual reformatting, then made some fixes.

IMHO, the reformatting makes the HTML usable -- the nesting is so broken, it's very difficult to make any changes. Code should be usable and modifiable. Github is for history. The HTML formatting currently is just horrible.

Plus...hidden in all that formatting mess is broken HTML. Which is what millions of users are experiencing currently. Why not serve correct, validated HTML instead?

I understand nervousness. But, this PR demonstrates that EVERY file now validates properly, which is a big deal. More testing leads to less nervousness. If this breaks something, then we should be testing that something.

bjones1:
Thanks! I appreciate it!

PR #1150, opened 2018-10-15
Rewrite new server test using pytest

bjones1:
I rewrote `test_server.py` to use [py.test](https://docs.pytest.org/en/latest/), a better testing framework. IMHO, we should migrate to this -- it does a much easier and better job of supporting testing.

### DIFF
--- a/requirements-test.txt
+++ b/requirements-test.txt
@@ -1,3 +1,4 @@
 coverage
 mock
 py_w3c
+pytest
--- a/tests/run_tests.py
+++ b/tests/run_tests.py
@@ -68,7 +68,7 @@ if __name__ == '__main__':
     with pushd('../../..'):
         # Now run tests.
         xqt('{} -m coverage erase'.format(sys.executable),
-            '{} -m unittest applications.runestone.tests.test_server'.format(sys.executable),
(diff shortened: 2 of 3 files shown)


bnmnetp:
Hey Bryan,

So I was going to accept this but wanted to try running everything locally first, so I branched and pulled this PR. and ran the tests and this is what I got:

```

bjones1:
It looks like you and I might have slightly different versions of web2py, which makes for fewer HTML validation errors on this page. (The errors come from HTML produced by web2py, not from our templates.) I'm running v. 2.17.1, and the Travis CI server runs whatever the latest version is.

bnmnetp:
OK, I'm at 2.16.1  -- I'll upgrade

bnmnetp:
Do you think we should we have Travis run a specific version?  Latest is good, but we may want to have another run where we are using the version that is live.  I definitely do not upgrade the live server every time web2py does a release.

bjones1:
I think running on a live version makes the more sense, rather than the latest. We can then manage any surprises that come from upgrading.

bjones1:
Thanks!

PR #1152, opened 2018-10-23
Add: Capability to use two-factor authentication.

bjones1:
**Background:** I use Runestone to give tests in my course. I use two-factor authentication, where only test proctors have the second password, to ensure that students can't log in from home to take the test. 

This PR adds the possibility to using two-factor authentication (but doesn't enable it without uncommenting two lines in `db.py`).

### DIFF
--- a/views/default/user.html
+++ b/views/default/user.html
@@ -98,36 +98,51 @@
             {{ if request.args(0) == 'login': }}
                 <div id="web2py_user_form" class="col-md-4">
                     <form class="form-horizontal" method="post">
-                        <div class="form-group">
-                            <div class="col-sm-offset-2">
-                                <label for="auth_user_username" class="control-label">Username</label>
-                                <input type="text" class="form-control col-xs-4" id="auth_user_username" name="username"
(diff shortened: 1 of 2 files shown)


views/default/user.html
```diff
@@ -98,36 +98,51 @@ <h4 class="col-md-4">{{=T( request.args(0).replace('_',' ').capitalize() )}}</h4
...
                             </div>
-                        </div>
-                        <div class="form-group">
-                            <div class="col-sm-offset-2 col-sm-4">
-                                <button type="submit" class="btn btn-default">Login</button>
+                            <div class="form-group">
+                                <div class="col-sm-offset-2 col-sm-4">
+                                    <button type="submit" class="btn btn-default">Login</button>
+                                </div>
                             </div>
```
bjones1: The github diff tool makes this look awful. It's actually just an added indent. Here's a snapshot from the SmartGit diff viewer that makes it clearer.

image

bjones1:
Thanks!

PR #1153, opened 2018-10-26
preview_question endpoint fixes

bjones1:
This PR fixes several bugs and weaknesses in the `ajax/preview_question` endpoint:

- It now works with Unicode.
- It doesn't produce internal server errors on bad params.
- It works on Window, Linux, and Mac.

There are also a few minor improvements to the test bench.

### DIFF
--- a/controllers/ajax.py
+++ b/controllers/ajax.py
@@ -920,9 +920,20 @@ def preview_question():
         env = dict(os.environ)
         # Prevent any changes to the database when building a preview question.
         del env['DBURL']
-        res = subprocess.call([sys.executable, '-m', 'runestone', 'build'], cwd='applications/{}/build/preview'.format(request.application), env=env)
-        if res != 0:
-            return json.dumps('Error: Runestone build failed.')
+        # Run a runestone build.
+        popen_obj = subprocess.Popen(
+            [sys.executable, '-m', 'runestone', 'build'],
+            # The build must be run from the directory containing a ``conf.py`` and all the needed support files.
+            cwd='applications/{}/build/preview'.format(request.application),
+            # Capture the build output in case of an error.
+            stdout=subprocess.PIPE, stderr=subprocess.PIPE,
+            # Pass the modified environment which doesn't contain ``DBURL``.
+            env=env)
+        stdout, stderr = popen_obj.communicate()
+        # If there was an error, return stdout and stderr from the build.
+        if popen_obj.returncode != 0:
+            return json.dumps('Error: Runestone build failed:\n\n' +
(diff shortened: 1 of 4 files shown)


bnmnetp:
when I try to test this on my dev machine (macos) I get the following:

screen shot 2018-10-31 at 3 31 01 pm

bjones1:
Hmm. I'll do more testing on my end.

bnmnetp:
Heisenbug... 

The returned result was Error: Runestone build failed . which does not create a component that has a [data-component] attribute so this little line 
```
opt.orig = jQuery(`#${whereDiv} [data-component]`)[0]
``` 
was failing...

When I added

```
        serr = open('/Users/bradleymiller/previewerr.log','w')
        sout = open('/Users/bradleymiller/previewout.log','w')        
```
and then updated the subprocess.call to send output there to help me debug it all worked

Not sure what to think...

bjones1:
So, if the reST is correct, the behavior of the PR matches the pre-PR behavior, correct? However, having better, more descriptive errors in the case of a reST syntax error would help. Of course, the JS will need to be updated at some point to display this more nicely.

For future reference: there are several types of errors the JS needs to deal with:

1. Errors that still produce build output.

``` HTML
<div class="system-message">
<p class="system-message-title">System Message: ERROR/3 (<tt class="docutils">F:\Runestone\web2py\applications\runestone\build\preview\_sources\index.rst</tt>, line 1)</p>
<p>On line 1, the last item in a fill-in-the-blank question must be a bulleted list.</p>
<div class="highlight-default notranslate"><div class="highlight"><pre><span></span><span class="o">..</span> <span class="n">fillintheblank</span><span class="p">::</span> <span class="n">testme</span>

  <span class="n">Testing</span><span class="o">.</span>

  <span class="o">-</span>  <span class="p">:</span><span class="n">a</span><span class="p">:</span> <span class="n">b</span>
    <span class="p">:</span><span class="n">c</span><span class="p">:</span> <span class="n">d</span>
</pre></div>
</div>
</div> 
```

2a. Not necessarily an error, but no Runestone component is created.

```
Error: Runestone content missing.
```

2b. Errors that produce a non-zero return value from the Runestone build. @bnmnetp, help! Do you have an example of some reST? I updated my code to catch this, but need a test case.

bjones1:
I've pushed an update to capture stdout and stderr if the runestone build exit code isn't 0.

bnmnetp:
No, correct rst was seemingly failing to build unless I added the stdout and stderr parameters.

I'll test again with your latest and see what happens.

bjones1:
Thanks!

PR #1155, opened 2018-11-02
Prevent changes to the userid from the profile page

bjones1:
Someone mentioned this on Slack a while ago. Here's a fix for it. The case: the profile endpoint allows logged-in users to modify their userid to any value; this PR prevents that from happening.

It also includes TODO items for thing to fix (changing e-mail address, e-mail validation, etc.)

### DIFF
--- a/controllers/default.py
+++ b/controllers/default.py
@@ -67,7 +67,6 @@ def user():
         form[0].insert(-1, my_extra_element)
         form.element('#auth_user_username')['_readonly'] = True
 
-    if 'profile' in request.args(0):
         form.vars.course_id = auth.user.course_name
         if form.process().accepted:
             # auth.user session object doesn't automatically update when the DB gets updated
--- a/controllers/default.py
+++ b/controllers/default.py
@@ -68,7 +68,10 @@ def user():
         form.element('#auth_user_username')['_readonly'] = True
 
         form.vars.course_id = auth.user.course_name
-        if form.process().accepted:
+        if form.validate():
+            # Prevent the username from being changed by deleting it before the update. See http://web2py.com/books/default/chapter/29/07/forms-and-validators#SQLFORM-without-database-IO.
+            del form.vars.username
+            form.record.update_record(**dict(form.vars))
             # auth.user session object doesn't automatically update when the DB gets updated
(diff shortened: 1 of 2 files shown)


controllers/default.py
```diff
@@ -67,9 +67,11 @@ def user():
         form[0].insert(-1, my_extra_element)
         form.element('#auth_user_username')['_readonly'] = True
 
-    if 'profile' in request.args(0):
```
bnmnetp: Hmm, I'm worried that this one has unintended consequences since this crazy controller gets called for registration and profile editing and all kinds of other pages.  The point of this if statement was to make sure that this processing only happened for profile updates, not everything.

I don't particularly like it but it was the recommended setup when we started using web2py.

bnmnetp: Also, users are not allowed to edit their username as it is.  So I'm not really clear on why this is necessary in the first place??

bjones1: It's necessary because we provide an endpoint that allows arbitrarily changing a user's userid, which could cause some "interesting" problems. The endpoint should not allow this; simply marking a HTML field as read-only isn't sufficient protection.

bjones1: In terms of unintended consequences, that's certainly possible. Steps I've taken to reduce this risk:

1. Testing. Additional tests in this PR do a fair amount of exercising of this endpoint.
2. Design. Note that line 55 of this source file contains the **exact same** statement. Why would `request.args` be changed?
3. Documentation. Notes following this change explain the code I'm using and why, which makes it easier to make changes in the future.

mlhess: With LTI integration, this could have some interesting side effects and should be fixed.  If we merge the other changes that I worked on we could flag this using the LTI_MODE (it might be called something else) 

However, being able to change your userid to anything, should not be allowed as it is not intended behavior.  

The form field  is marked readonly but the backend does not honor that, relying on front end validation only is an issue. 

If we don't want to fix this, we should remove the flag on the username field to be readonly.

bnmnetp: So this is a workaround for a bug in web2py ??

bjones1: If grades are involved, students can be surprisingly resourceful...

PR #1156, opened 2018-11-02
Fix: Clean up the section name in the default/user/profile file.

bjones1:
A typical entry for another field in the `default/user/profile` form:
``` HTML
<tr id="auth_user_first_name__row">
    <td class="w2p_fl">
        <label class="" for="auth_user_first_name" id="auth_user_first_name__label">First Name: </label>
    </td>
    <td class="w2p_fw">
        <input class="string" id="auth_user_first_name" name="first_name" type="text" value="test" />
    </td>
    <td class="w2p_fc"></td>
</tr>
```

Before this PR, the "Section Name" looked like this:
``` HTML
<tr>
    <td>
        <label>Section Name</label>
    </td>
    <td>
        <input name="section" type="text" value="default" />
    </td>
</tr>
```

After this PR, it matches other fields:
``` HTML
<tr id="auth_user_section__row">
    <td class="w2p_fl">
        <label for="auth_user_section" id="auth_user_section__label">Section Name: </label>
    </td>
    <td class="w2p_fw">
        <input class="string" id="auth_user_section" name="section" type="text" value="default" />
    </td>
    <td class="w2p_fc"></td>
</tr>
```

### DIFF
--- a/controllers/default.py
+++ b/controllers/default.py
@@ -5,7 +5,9 @@ import requests
 from urllib import unquote
 from urllib2 import HTTPError
 import logging
+
 from gluon.restricted import RestrictedError
+from gluon.tools import addrow
 
 logger = logging.getLogger(settings.logger)
 logger.setLevel(settings.log_level)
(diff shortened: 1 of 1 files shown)


bjones1:
Thanks!

PR #1159, opened 2018-11-08
sanitize answers to fitb to prevent script injection when user clicks…

presnick:
… on 'check me'

### DIFF
--- a/controllers/ajax.py
+++ b/controllers/ajax.py
@@ -9,6 +9,7 @@ import os
 import sys
 from io import open
 from lxml import html
+import bleach
 
 logger = logging.getLogger(settings.logger)
 logger.setLevel(settings.log_level)
@@ -100,8 +101,8 @@ def hsblog():
         #       (db.fitb_answers.div_id == div_id) &
         #       (db.fitb_answers.course_name == auth.user.course_name) &
         #       (db.fitb_answers.correct == 'T')).count() == 0:
-            answer = request.vars.answer
-            correct = request.vars.correct
+            answer = bleach.clean(request.vars.answer)
+            correct = bleach.clean(request.vars.correct)
             db.fitb_answers.insert(sid=sid, timestamp=ts, div_id=div_id, answer=answer, correct=correct, course_name=course)
 
     elif event == "dragNdrop" and auth.user:
--- a/requirements.txt
+++ b/requirements.txt
@@ -15,4 +15,5 @@ sphinxcontrib-paverutils>=1.17
 SQLAlchemy>=1.0.14
 cssselect>=1.0
 oauth2>=1.9
+bleach>=3.0.2
 -e rsmanage/


bjones1:
I think this is failing because of the version of web2py used for testing, not because of these changes. We need to download a known web2py version, instead of whatever the latest and greatest is. (These tests were written for web2py 2.17.1, but it's running against 2.17.2, which evidently breaks some of the HTML.)

controllers/ajax.py
```diff
@@ -100,8 +101,8 @@ def hsblog():
         #       (db.fitb_answers.div_id == div_id) &
         #       (db.fitb_answers.course_name == auth.user.course_name) &
         #       (db.fitb_answers.correct == 'T')).count() == 0:
-            answer = request.vars.answer
-            correct = request.vars.correct
+            answer = bleach.clean(request.vars.answer)
+            correct = bleach.clean(request.vars.correct)
```
bjones1: Why bleach correct? It's just a True or False value.

controllers/ajax.py
```diff
@@ -100,8 +101,8 @@ def hsblog():
         #       (db.fitb_answers.div_id == div_id) &
         #       (db.fitb_answers.course_name == auth.user.course_name) &
         #       (db.fitb_answers.correct == 'T')).count() == 0:
-            answer = request.vars.answer
-            correct = request.vars.correct
+            answer = bleach.clean(request.vars.answer)
```
bjones1: What if the question required an answer such as `<div>` or some other HTML tag (a course on web design, for example)? It makes more sense to me to escape any HTML entities, probably using `cgi.escape`.

bnmnetp: How exactly would script injection come into play on a fill in the blank?  wouldn't we have to eval the answer or redisplay the answer in order for that to occur?  Simply entering a script into the text input box doesn't do it.  Maybe I'm a really bad hacker but I didn't get it to inject anything as is.

bjones1: Correct, redisplaying it (from the grading interface) is where I see the vulnerability. @mlhess can probably provide more info.

bjones1: And hijacking the grading interface gives the malicious user the ability to changes grades and act as an admin...

bnmnetp: Ok, we'll call this one an ounce of prevention.  Those Michigan students...

bjones1: Before this can be merged, RunestoneComponents need to be updated -- all correct answers should also be passed through bleach/escape/whatever is chosen for the sanitizer. Also, this needs tests for exactly this case!

mlhess: > Ok, we'll call this one an ounce of prevention. Those Michigan students...

This is not about students, but about third parties who see this as an easy way to exploit. The students would be the victim.

bnmnetp: I'm not following you @bjones1 -- What needs to be updated in components?  This seems like a very straightforward change.

bnmnetp: Wouldn't it be simpler to just make sure that fitb answers are escaped before they are displayed in the grading interface?  -- This would also apply to the short answer directive as well since that allows for free input by a user.

mlhess: This gets into the debate on filter on output vs filter on input. 

This is a good reference for that: http://acko.net/blog/safe-string-theory-for-the-web/

bjones1: The quote from that reference, "You should aim to keep data in the most natural form possible, until right before output." I like the idea; that makes sense. However, I'm also afraid that we'll sanitize all output except the one that really matters -- there can be multiple outputs to track down, but there's a single input source.

So, I'm not sure which is better. However, doing nothing is clearly worse :).

bnmnetp: Guys,

1.  In the grading interface and elsewhere the stored answer is restored for viewing by setting the "value" attribute on the text input box.  The answer is not rendered as html, and I haven't been able to find a description of how to make script injection work in this context.  Please forward me a link if I'm wrong here.

2.  The answer is stored in two places.  localStorage in the browser as well as in the database.  Which in my mind makes the only real solution (I'm not sure its even a problem given that we are assigning a string to an attribute) to do any escaping at display time.

I think this whole PR is a red herring.

Brad

bjones1: That makes sense -- per @mlhess's article, a good approach is filter on output and that's already taken care of (as you point out). There's no other place fitb answers get shown? (Perhaps somewhere in the student progress tab in one of the histograms / details you can click on?)

presnick:
This request deprecated in favor of cleaning on output, implemented in https://github.com/RunestoneInteractive/RunestoneServer/commit/2a9edc2787374f4049d4d8777f49d07816ba63ce

bjones1:
I'd suggest using [cgi.escape](https://docs.python.org/2/library/cgi.html#cgi.escape) instead of `bleach` to preserve the correct display of these items.

mlhess:
I don't think that removes all XSS
https://bugs.python.org/issue26398

bjones1:
Per the bug report, `cgi.escape` doesn't (by default) escape double quotes, so (as I understand it) it should not be used for assigning to an HTML attribute without specifying `quote=True`. The underlying question is: in what context will these strings be used? As raw HTML? As an attribute? Inside a `<script>` tag? Looking at the code, I see Bleach used in the following places:

1. In [_getStudentResults](https://github.com/RunestoneInteractive/RunestoneServer/blob/master/controllers/ajax.py#L616). This is called by:
    - [gettop10answers](https://github.com/RunestoneInteractive/RunestoneServer/blob/master/controllers/ajax.py#L745), which returns the data as JSON to [fitb.js](https://github.com/RunestoneInteractive/RunestoneComponents/blob/master/runestone/fitb/js/fitb.js#L335).
        - The [compareFITB](https://github.com/RunestoneInteractive/RunestoneComponents/blob/master/runestone/fitb/js/fitb.js#L335) callback then includes the data in an HTML table.
    - [getaggregateresults](https://github.com/RunestoneInteractive/RunestoneServer/blob/master/controllers/ajax.py#L691), which returns the data as JSON to [mchoice.js](https://github.com/RunestoneInteractive/RunestoneComponents/blob/master/runestone/assess/js/mchoice.js#L557).
        - The [compareModal](https://github.com/RunestoneInteractive/RunestoneComponents/blob/master/runestone/assess/js/mchoice.js#L501) callback then retrieves this as [datadict.reslist](https://github.com/RunestoneInteractive/RunestoneComponents/blob/master/runestone/assess/js/mchoice.js#L532) and calls [instructorMchoiceModal](https://github.com/RunestoneInteractive/RunestoneComponents/blob/master/runestone/assess/js/mchoice.js#L489). This function includes the data as part of an HTML table.
2. In [getaggregateresults](https://github.com/RunestoneInteractive/RunestoneServer/blob/master/controllers/ajax.py#L659). I honestly don't understand how this is used and even if it makes it back to the client in some form.

So, from item 1 above, we need to escape for inclusion in HTML, so `cgi.escape` is the correct choice. Since item 2 above is difficult to follow, I can't say what's appropriate. However, bleach clearly removes potentially correct pieces of an answer.

PR #1161, opened 2018-11-15
Define web2py version for testing

bjones1:
This PR builds on #1160 by explicitly specifying which version of web2py to use for testing.


bnmnetp:
@bjones1 At a quick glance this one looks like it includes all the changes from #1160 ??

bjones1:
Correct. Until #1160 is merged, everything will fail. So, it should be first on your list, then other PRs.

PR #1168, opened 2018-12-05
Registration fix

bjones1:
If a user makes a data entry error when registering (for example, the passwords don't match), then the course name they entered is replaced by the course ID from the database when the form is re-displayed with an error. This fix keeps the course name from being changed to a course ID. It also adds tests for this case.


bnmnetp:
Nice, this has been an annoyance for a while.

bjones1:
Thanks!

PR #1169, opened 2018-12-05
Server side fitb

bjones1:
This PR allows grading of fill-in-the-blank questions on the server. However, server-side grading only occurs when a course is built with server-side grading enabled. Therefore, current behavior (client-side grading) will be maintained until a RunestoneComponents PR provides the necessary updates.

Merge #1168 before this.

### DIFF
--- a/modules/rs_grading.py
+++ b/modules/rs_grading.py
@@ -2,15 +2,12 @@ import datetime
 import logging
 from math import ceil
 from psycopg2 import IntegrityError
+
+from gluon import current
 from outcome_request import OutcomeRequest
 
-# When testing, the ``settings`` object isn't defined. Import it in this case.
-try:
-    logger
-except:
-    execfile('applications/runestone/models/0.py', globals())
-logger = logging.getLogger(settings.logger)
-logger.setLevel(settings.log_level)
+logger = logging.getLogger(current.settings.logger)
+logger.setLevel(current.settings.log_level)
 
 def _profile(start, msg):
     delta = datetime.datetime.now() - start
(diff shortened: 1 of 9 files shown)


bjones1:
Fixed now. I've written lots of tests for this, but they won't pass until a RunestoneComponents PR is merged.

modules/rs_grading.py
```diff
@@ -2,15 +2,12 @@
...
 
-# When testing, the ``settings`` object isn't defined. Import it in this case.
-try:
-    logger
-except:
-    execfile('applications/runestone/models/0.py', globals())
-logger = logging.getLogger(settings.logger)
-logger.setLevel(settings.log_level)
+logger = logging.getLogger(current.settings.logger)
+logger.setLevel(current.settings.log_level)
```
bnmnetp: We should make a todo  that functions like _scorable_mchoice_answers no longer need to pass db as a parameter.

bjones1: Good point. Do you want me to submit an issue? Or would you like to?

bjones1:
That's correct -- unless server-side grading is explicitly enabled for a book by setting `runestone_server_side_grading` to `True` and all required conditions are met, this will do all grading client-side. The conditions are:

- The use of Runestone services (``eBookConfig.useRunestoneServices === true``)
- A valid login (``eBookConfig.loginRequired === true``)
- Logging enabled (``eBookConfig.logLevel  > 0``)

bnmnetp:
I hear you.  I'll finish up my testing on this over the weekend.

bjones1:
Great, thanks!

bjones1:
Thanks!

PR #1173, opened 2018-12-28
Workaround: Never grade all problems for all students

bnmnetp:
@presnick , @bjones1 

This change was inspired by a hack published by one of our professors.  

It subverts grading ALL problems for ALL students when just the assignment is selected.  The underlying problem is that autograde is too slow for large classes or large assignments, but this gets around the problem by grading one student at a time.

I think its better than letting the majority of users suffer.

Comments?

### DIFF
--- a/static/js/admin.js
+++ b/static/js/admin.js
@@ -155,16 +155,19 @@ function autoGrade() {
     }
 
     if (assignment != null && question === null && studentID == null) {
+        let i = 0;
         for (let s in students) {
             params.data.sid = s;
             params.async = false;
             let res = jQuery.ajax(params);
-            $("#autogradingform").append(`${res.responseJSON.message} for ${s} </br>`)
+            setTimeout(function(s) {
+                $("#autogradingform").append(`${res.responseJSON.message} for ${s} </br>`)
+            }, 10, s);
         }
         calculateTotals();
         $("#autogradesubmit").prop("disabled", false);
     } else {
-        jquery.axax(params).always(function () {
+        jQuery.ajax(params).always(function () {
             calculateTotals();
             $("#autogradesubmit").prop("disabled", false);
         });
--- a/static/js/admin.js
+++ b/static/js/admin.js
@@ -138,7 +138,7 @@ function autoGrade() {
     var question = getSelectedItem("question")
     var studentID = getSelectedItem("student")
     var enforceDeadline = $('#enforceDeadline').is(':checked')
-    jQuery.ajax({
+    var params = {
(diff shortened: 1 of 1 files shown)


bjones1:
Oh, aargh. I hate kludges. They become maintenance burdens. Have either of yall had time to profile the grader? Is this an easy fix, or something hard to fix?

bnmnetp:
1.  Yes, I have profiled the grader and it is not easy to fix.

2.  I could make an argument that from a UX perspective this is really an improvement as the instructor gets to see progress while they wait.

3.  The other immediate alternative would be to move grading to a scheduled task -- like building/rebuilding a course.  But that is more work than this.

Right now autograding is just outright unusable for anyone with a reasonably large class.  I'd like to make it usable.

bnmnetp:
I would also just emphasize that this only changes admin.js no server side changes are needed.

static/js/admin.js
```diff
@@ -155,16 +155,19 @@ function autoGrade() {
     }
 
     if (assignment != null && question === null && studentID == null) {
+        let i = 0;
         for (let s in students) {
```
bjones1: Is `students` an array of student names? (Aargh, the grading interface doesn't user userids!) I assume there's no easy way around this for now?

static/js/admin.js
```diff
@@ -155,16 +155,19 @@ function autoGrade() {
...
             params.data.sid = s;
             params.async = false;
             let res = jQuery.ajax(params);
-            $("#autogradingform").append(`${res.responseJSON.message} for ${s} </br>`)
+            setTimeout(function(s) {
+                $("#autogradingform").append(`${res.responseJSON.message} for ${s} </br>`)
+            }, 10, s);
```
bjones1: Does this imply that all autograde messages will appear no sooner than 10 seconds after the grading starts?

bjones1:
OK, thanks. Given a hard fix, your workaround seems like the best short-term approach.

On a side note, we really need to fix the whole client-side grading interface. The use of first name, last name to identify users causes a lot of problems, and there are several bugs I've seen while doing some grading.

static/js/admin.js
bnmnetp: These are usernames/sids

static/js/admin.js
bnmnetp: 10 milliseconds

bnmnetp:
I don’t believe we are using first name and last name for anything more than display. But please correct me if I’m wrong.

PR #1182, opened 2019-01-05
fixes for lti practice grading

presnick:


controllers/assignments.py
```diff
@@ -632,9 +632,8 @@ def chooseAssignment():
 def _get_lti_record(oauth_consumer_key):
     return db(db.lti_keys.consumer == oauth_consumer_key).select().first()
 
-def _get_course_practice_record(course_name, user_id):
-    return db((db.course_practice.course_name == course_name) &
-              (db.course_practice.auth_user_id == user_id)).select().first()
+def _get_course_practice_record(course_name):
```
bnmnetp: Hmmm, why leave off the user id?  Wouldn't this just return the first random record among all students in the class?

presnick: This is actually the info about practice settings for *the course*.  The auth_user_id field is the id of the instructor.  We were failing to get the course settings when a student's user_id was passed.

The table probably shouldn't have an auth_user_id field at all-- Iman uses that to make sure that when he randomizes courses to different conditions, all the courses for an instructor get the same condition. He could do that with an extra table join instead.

presnick: @bnmnetp I suggest you merge this and, if you want, make a request to @ImanYZ  to get rid of the auth_user_id field from the course_practice table.

PR #1183, opened 2019-01-06
Feature server modes

presnick:
This feature introduces three settings variables that can be configured in 1.py. By default, they are set in 0.py to yield an installation that mimics what runestoneAcademy currently runs.

When academy_mode is off, some academy-specific stuff disappears (e.g., running a scheduler, allowing creation of new courses, showing carousel).

When lti_only mode is turned on, registration and login links and pages are disabled.

When coursera_mode is turned on, some of the feedback text changes.


bjones1:
@presnick, would you be interested in developing any unit tests for this? If so, I'd be happy to help -- I've been working to improve the test framework, and this might be a good chance to see if it works for others, not just me.

presnick:
A good idea in principle. My time is very limited right now, though. If you can get me started, I may be able to finish it.

bjones1:
Great! I'll work on it a bit to get started with some tests, then let you know.

views/assignments/doAssignment.html
```diff
@@ -31,7 +28,7 @@
 <script type="text/javascript" src="/{{=request.application}}/static/{{=course_id}}/_static/jquery.ba-bbq.min.js"></script>
 <script type="text/javascript" src="/{{=request.application}}/static/{{=course_id}}/_static/jquery.jsPlumb-1.3.10-all-min.js"></script>
 <script type="text/javascript" src="/{{=request.application}}/static/{{=course_id}}/_static/d3.v2.min.js"></script>
-
+<script type="text/javascript" src="{{=URL('static', 'js/selfgrade.js')}}"></script>
```
bnmnetp: What and where is selfgrade.js ?? I don't see a new file for this added to the PR and I can't find it anywhere in the existing code base.

Should this be conditionally included only for coursera mode or something?  It doesn't seem like it needs to be included for academy mode.

views/admin/index.html
```diff
@@ -110,6 +110,8 @@ <h3 style="color: orange;">Beta - New Instructor Interface!</h3>
 </div>
 
 <div class="col-md-6">
+
+    {{ if settings.academy_mode: }}
```
bnmnetp: Can we get the indentation levels right?  The start and ending indentations of this and the pass at the end don't match.  Plus the if is indented but none of the html inside the if is indented at all.  This makes it super confusing to read and follow.

bnmnetp: pinging @presnick I will accept and get this onto master if you can clean up the issues I identified.

bnmnetp:
@presnick -- I would like to get this merged before it gets any more stale.  Can we please deal with the missing selfgrade.js file??

presnick:
I added the missing static/js/selfgrade.js

Sorry about that.

presnick:
I also merged in the latest master, so it least the diffs are clearer...

bnmnetp:
@presnick doing carful testing of this today.  However the doAssignment page is broken.  The template is looking for autgradingUrl and probably calcTotalsURL however they are not provided by the controller.  I'm guessing that lines 63,64 should only be included for coursera mode ??

bnmnetp:
I made the change to only include them in coursera mode, so no need for you to do it if that is the right solution.

presnick:
Yes, that’s a fine plan. Thanks for fixing. The self-autograde feature was
hacked together in a hurry and is only needed for Coursera for now. Some
more general version may be forthcoming in the future.

PR #1184, opened 2019-01-07
Add: Server-side tests of fitb.

bjones1:
This is a simple PR that adds much more testing for the server-side fill-in-the-blanks question type. The tests demonstrate that grading works properly on the server.

### DIFF
--- a/requirements-test.txt
+++ b/requirements-test.txt
@@ -3,3 +3,5 @@ mock
 py_w3c
 pytest
 contextlib2
+# Install the development version of the Runstone Components, rather than the released version.
+https://github.com/RunestoneInteractive/RunestoneComponents/archive/master.zip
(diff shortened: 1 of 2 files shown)


bjones1:
Tests fail since they're running against the currently released version of RunestoneComponents, not the current Github version. @bnmnetp, what do you think about changing the test suite to use the Github version of RunestoneComponents instead?

bnmnetp:
@bjones1 using the github version is probably better as we so often are doing development that requires changes on both sides.

bjones1:
Sounds good. I'll create a separate PR for that, then rebase this on top of it.

bjones1:
I just included it here, since the two go together. This is ready to merge.

bjones1:
Thanks!

PR #1189, opened 2019-01-12
better display of readings in reading assignments

presnick:
If no points for that chapter, don't show points earned line

Show full chapter and subchapter name rather than the compressed labels.

### DIFF
--- a/controllers/assignments.py
+++ b/controllers/assignments.py
@@ -510,7 +510,11 @@ def doAssignment():
             return redirect(URL('assignments', 'chooseAssignment'))
 
     questions = db((db.assignment_questions.assignment_id == assignment.id) & \
-                   (db.assignment_questions.question_id == db.questions.id)) \
+                   (db.assignment_questions.question_id == db.questions.id) & \
+                   (db.chapters.chapter_label == db.questions.chapter) & \
+                   (db.chapters.course_id == course.course_name) & \
+                   (db.sub_chapters.chapter_id == db.chapters.id) & \
+                   (db.sub_chapters.sub_chapter_label == db.questions.subchapter)) \
         .select(db.questions.name,
                 db.questions.htmlsrc,
                 db.questions.id,
@@ -519,8 +523,9 @@ def doAssignment():
                 db.assignment_questions.points,
                 db.assignment_questions.activities_required,
                 db.assignment_questions.reading_assignment,
+                db.chapters.chapter_name,
+                db.sub_chapters.sub_chapter_name,
                 orderby=db.assignment_questions.sorting_priority)
-
 
     try:
         db.useinfo.insert(sid=auth.user.username,act='viewassignment',div_id=assignment.name,
@@ -564,15 +569,18 @@ def doAssignment():
             comment=comment,
             chapter=q.questions.chapter,
             subchapter=q.questions.subchapter,
+            chapter_name=q.chapters.chapter_name,
+            subchapter_name=q.sub_chapters.sub_chapter_name,
             name=q.questions.name,
             activities_required=q.assignment_questions.activities_required
         )
         if q.assignment_questions.reading_assignment:
             # add to readings
-            if q.questions.chapter not in readings:
+            ch_name = q.chapters.chapter_name
+            if ch_name not in readings:
                 # add chapter info
                 completion = db((db.user_chapter_progress.user_id == auth.user.id) & \
-                                (db.user_chapter_progress.chapter_id == q.questions.chapter)).select().first()
+                                (db.user_chapter_progress.chapter_id == ch_name)).select().first()
                 if not completion:
                     status = 'notstarted'
                 elif completion.status == 1:
@@ -581,7 +589,7 @@ def doAssignment():
                     status = 'started'
                 else:
                     status = 'notstarted'
-                readings[q.questions.chapter] = dict(status=status, subchapters=[])
+                readings[ch_name] = dict(status=status, subchapters=[])
 
             # add subchapter info
             # add completion status to info
@@ -598,7 +606,7 @@ def doAssignment():
                 status = 'notstarted'
             info['status'] = status
 
-            readings[q.questions.chapter]['subchapters'].append(info)
+            readings[ch_name]['subchapters'].append(info)
             readings_score += info['score']
 
         else:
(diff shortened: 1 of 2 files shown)


controllers/assignments.py
```diff
@@ -510,7 +510,11 @@ def doAssignment():
             return redirect(URL('assignments', 'chooseAssignment'))
 
     questions = db((db.assignment_questions.assignment_id == assignment.id) & \
-                   (db.assignment_questions.question_id == db.questions.id)) \
+                   (db.assignment_questions.question_id == db.questions.id) & \
```
bnmnetp: @presnick what is the purpose of all these extra conditions?  This is causing some people problems where custom questions they have addedto their assignments are suddenly not showing up.

presnick: The purpose is to get the chapter name and subchapter name for readings, in order to display them rather than the no-spaces version that was displaying before.

Is the problem that questions not associated with a chapter (i.e., authored through the interface instead of an rst file) are getting filtered out by these inner joins?

One solution, though not as efficient, might be to do a separate query for each of the questions that is of type reading, to get its subchapter name.

I think it would be better to do a left outer join. We'd have to use the different web2py query syntax , as described here: http://www.web2py.com/books/default/chapter/29/06/the-database-abstraction-layer#One-to-many-relation

presnick: I'm trying to compose and test a query right now.

bnmnetp: Exercises created through the web interface are stored in the sub chapter called Exercises.  We made a convention that All chapters must have an Exercises sub chapter.

So, the problem is that the StudentCSP book does not follow the convention of calling the Exercises chapter Exercises.  @ericsonga contacted me today because a bunch of her custom problems suddenly disappeared.  I couldn’t re-create the problem in any of the books I use, but when I dug deeper I saw that her book doesn’t follow the convention.  

Relying on everyone to follow conventions is risky, and this demonstrates that.

There should be a better way to ensure that every chapter has a sub chapter called Exercises.  Or else we should have a second drop down in the exercise authoring interface to make the author choose a sub chapter as well as the chapter.

presnick: Well, I've created the following alternative query. I've tested it on my local machine and I am able to get a custom question to show up on an assignment page.

HOWEVER: after doing that, I also checked that, even with my previous query, I also go the same custom question to show up on an assignment page. So, not having reproduced the error, I'm not sure if this will fix it.

`    questions = db((db.chapters.course_id == course.course_name) & \
                   (db.sub_chapters.chapter_id == db.chapters.id) & \
                   (db.assignment_questions.assignment_id == assignment.id)) \
        .select(db.questions.name,
                db.questions.htmlsrc,
                db.questions.id,
                db.questions.chapter,
                db.questions.subchapter,
                db.assignment_questions.points,
                db.assignment_questions.activities_required,
                db.assignment_questions.reading_assignment,
                db.chapters.chapter_name,
                db.sub_chapters.sub_chapter_name,
                orderby=db.assignment_questions.sorting_priority,
                join=db.assignment_questions.on(db.questions.id == db.assignment_questions.question_id),
                left=[db.chapters.on(db.questions.chapter == db.chapters.chapter_label),
                      db.sub_chapters.on(db.questions.subchapter == db.sub_chapters.sub_chapter_label)]
                )`

bnmnetp:
Paul,

I'm pretty sure that won't fix the problem.  The issue is that when you SAVE an exercise it saves it with the subchapter of Exercises  but Barbs book doesn't have any subchapters called Exercises.  So the join will always end up excluding those questions.

I've asked Barb to update her book so that it follows our convention and then your code will work for her book with out modification.

This just shows how complicated and interconnected all of our code and books are...  Even a good set of Unit tests would probably not have caught this one.

Brad

presnick:
Maybe if we moved this clause down into the left joins, it would work for that case?

`(db.sub_chapters.chapter_id == db.chapters.id)`

controllers/assignments.py
bnmnetp: I think I need to read the docs on what it means to use the join parameter  inside select the select method.

presnick: Here's a sample of a query generated from that, in case it helps.

```
SELECT  questions.name, questions.htmlsrc, questions.id, questions.chapter, questions.subchapter, assignment_questions.points, assignment_questions.activities_required, assignment_questions.reading_assignment, chapters.chapter_name, sub_chapters.sub_chapter_name 
FROM questions 
JOIN assignment_questions ON (questions.id = assignment_questions.question_id) 
LEFT JOIN chapters ON (questions.chapter = chapters.chapter_label) 
LEFT JOIN sub_chapters ON (questions.subchapter = sub_chapters.sub_chapter_label) 
WHERE (((chapters.course_id = 'fopp') 
AND (sub_chapters.chapter_id = chapters.id)) 
AND (assignment_questions.assignment_id = 5)
AND assignment_questions.reading_assignment IS NULL) 
ORDER BY assignment_questions.sorting_priority;
```

bjones1:
FYI, I just updated to use this and my assignments also disappeared. Is there a workaround, or something I"m missing in my book? I have a subchapter named Exercises in all chapters, and I don't use any custom problems.

For now, I've reverted this commit in my repo.

bnmnetp:
Hmmm, The only place I've run into trouble with this is in chapters without and `Exercises.rst` file capitalization is important.

Does your book use a `toctree.rst` file for each chapter?

bnmnetp:
@bjones1 - It is the file name that is important not the title of the section.  The title can be anything you want, but the subchapter_id comes from the name of the file.

bjones1:
Each subdirectory contains a `toctree.rst` and an `exercsises.rst`. Is that correct? Or should it be `Exercises.rst`?

bnmnetp:
Yes, capitalization is important, so it should be `Exercises.rst` not `exercises.rst`

bjones1:
OK, thanks -- that fixed it.

I would be nice to have some sort of warning during a build if this wasn't present. Everything else worked as far as I could tell until this commit.

bnmnetp:
Yes, I agree that would be a nice little addition to runestone build.  It would probably not be that hard either as we do a fairly systematic chapter by chapter writing of chapter/subchapter info.

PR #1197, opened 2019-01-23
Fix: Avoid exception when displaying the gradebook.

bjones1:
Sometimes, `score` is `None`, causing an exception.


bnmnetp:
Interesting how the order of operations works with this.  Not easy to read.

bjones1:
I agree. Adding parens would be nice.

PR #1200, opened 2019-01-24
LTI cleanup

bjones1:
Clean: 
- Remove unused variables, or unnecessary declarations of variables.
- Simplify control flow (no else after a return).

Fix:
- Use the provided URL unless ``settings.lti_uri`` is defined.


bjones1:
My pleasure!

PR #1201, opened 2019-01-27
URGENT; fix bug when there are 0 points for the assignment

presnick:
Bug introduced recently that may affect any teachers who create ungraded assignments.


bjones1:
Oops, thanks for this fix. Isn't it missing an `</h2>`?

presnick:
Restored the missing </h2>

PR #1203, opened 2019-02-06
Add: An option to charge for students to register for a course.

bjones1:


bjones1:
@bnmnetp, pinging. I think your recent pushes produced merge conflicts, which are very painful to solve for `runestone_test.sql`. Do you have any time to review this?

bnmnetp:
I added a single line to runestone_test.sql because I added a column to the courses table.   So it should not have created a big problem.

I'll try to find some time today or tonight to have a look.

bnmnetp:
Yes, they are very simple conflicts.  In both cases we added a column to the courses table, which caused conflicts in db.py and runestone_test.sql.   Very simple to merge those two changes.

bjones1:
OK, I'll take a look.

bjones1:
@bnmnetp, pinging. Do you have time to review this?

bnmnetp:
The only thing I'm a little worried about with this one is making the donate page unavoidable in between registering and the homepage of the class.   I mainly worry about this for high schools, but if it turns out to be an issue we can always add a flag for high school books so that this page is skipped.

Maybe this will increase the percentage of registered users who donate...

bjones1:
Thanks for reviewing and merging this and several others PRs. I appreciate it. The merges help me in my development process, since I'm still quite a few commits ahead of master, and I'd like to share my improvements with the community.

I can see your concern about making the donation page more obvious, but IMHO users are very good at not paying for things :). I think an institutional/paid model would be very appropriate, personally. Funding helps motivate development and produces a better product for everyone.

PR #1212, opened 2019-03-12
Fix: Let application name determine masterapp for session storage.

bjones1:
This is one more place where we can avoid hard-coding the application name, without changing the behavior of the Runestone server.


bjones1:
Thanks!

PR #1215, opened 2019-03-12
Fix: Display progress while grading without blocking.

bjones1:
This uses JavaScript's async capabilities to avoid blocking while performing grading. Without this, my browser (Chrome) freezes while doing grading. It also provide a nice display of what's currently being graded. With 180 students in my course, this is essential.

### DIFF
--- a/static/js/admin.js
+++ b/static/js/admin.js
@@ -155,17 +155,20 @@ function autoGrade() {
     }
 
     if (assignment != null && question === null && studentID == null) {
-        let i = 0;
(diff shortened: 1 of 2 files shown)


bnmnetp:
I'm don't fully understand why this one works for you while the previous solution didn't, but I'm happy to have it.

bjones1:
Thanks. Perhaps the longer delay (about 2 seconds per request in my class of 180 students) means the browser becomes less responsive, since JS blocks it?

PR #1217, opened 2019-03-14
LTI fixes

bjones1:
This contains some clean-up and bug fixes for the LTI interface.

One remaining item to resolve: at https://github.com/RunestoneInteractive/RunestoneServer/blob/master/lti.config#L17, there's supposedly a `course_id` and `course_name` custom parameter; however, https://github.com/RunestoneInteractive/RunestoneServer/blob/master/controllers/lti.py#L107 calls this `custom_course_id` and makes no mention of  `custom_course_name`, but does have a `custom_section_id`. So, I suspect that these names should be changed to agree. But, I'm not an LTI person and perhaps there's a nuance of the spec I don't understand.

### DIFF
--- a/controllers/lti.py
+++ b/controllers/lti.py
@@ -49,7 +49,7 @@ def index():
         userinfo = dict()
         userinfo['first_name'] = first_name
         userinfo['last_name'] = last_name
-        # In the `Canvas Student View <https://community.canvaslms.com/docs/DOC-13122-415261153>` as of 7-Jan-2019, the ``lis_person_contact_email_primary`` is an empty string. In this case, use the userid instead.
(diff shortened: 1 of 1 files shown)


bjones1:
Paul, thanks for your review. I always appreciate someone else's perspective on my code.

Yes, I run this on my production server. It's one class with ~180 students -- certainly not the size of Brad's server, but enough to show bugs and problems. All students log in through LTI, and I added the instructor fix so my TAs would automatically be given access.

The reason for the movement is based on how courses are marked as paid: an entry in `user_courses` for that user/course combo means paid. So, by putting the instructor code first, I can add this row before the user code check to see if it is there and request payment if not.

One thing I strongly dislike in this modification is the complete lack of testing. Do you know of any way to mock up an LTI testing environment? If I knew a bit more about what parameters to pass, I could work on tests, but I don't know enough OAuth and haven't really looked at the code enough to understand it. Would you be willing to advise me in this?

bjones1:
@bnmnetp. pinging...

bjones1:
Thanks!

PR #1220, opened 2019-04-09
Pivot reports

bnmnetp:
This  is in response to one of the top requested features I get from instructors, more granularity than they get in the student progress page and the ability to download data in a useful form. I think this addresses that without just giving them a dump of the firehose that is useinfo.

It gives them 4 reports

1.  student by subchapter with a count of the number of distinct activities each student has done.
2. student by subchapter-div_id with a timestamp of the first time they interacted with an activity
3. student by subchapter-div_id with a timestamp of the last time they interacted with an activity
5. student by subchapter-div_id with a count of the number of times they interactive with an activity

Screen Shot 2019-04-09 at 10 45 05 AM

To anticipate -- I'm thinking about how I would write a unit test for this @bjones1 .

### DIFF
--- a/controllers/dashboard.py
+++ b/controllers/dashboard.py
@@ -311,3 +311,37 @@ def exercisemetrics():
             })
 
     return dict(course_name=auth.user.course_name, course_id=auth.user.course_name, answers=answers, response_frequency=response_frequency, attempt_histogram=attempt_histogram, exercise_label=problem_metric.problem_text)
+
+import pandas as pd
+
+@auth.requires_login()
+def subchapoverview():
+    #course = db(db.courses.id == auth.user.course_id).select().first()
+    course = auth.user.course_name
+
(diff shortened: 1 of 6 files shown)


bnmnetp:
@bjones1 ,

I've added a very minimal unittest for this new controller.  The problem is that I used the "old" method and relied on data that is in the runestone_test database.

I would really like to change this and make it a pytest unit test as a learning exercise.  But here is the list of things that I would need to do.  I know some of them you have provisions for in your pytest library, but it would sure help if some of these things were documented.  As it stands I feel like it would take me longer to write the tests than it did to write the controller.

1. create a class
2. populate at least the questions table for that class
3. create an instructor for the class
4. create some students for the class
5. create an instructor for the class
6. Populate some activities into the useinfo table for those students.
7. Then **finally** call the controller and check the output of the table produced.

PR #1224, opened 2019-04-23
Fixes to get all tests running under Python3.7

bnmnetp:
@bjones1 These are the changes I made today to get all tests to pass under python-3.7.3

With the exception of the change I made to my logging.conf file, which is not tracked but was causing me loads of errors at first.

- differences in rounding were causing 4 of the autograding tests to fail
- My use of StringIO for my last PR was not python 3 compatible.


bnmnetp:
grrrrr:  Build is failing because of a version conflict.

```
pkg_resources.ContextualVersionConflict: (urllib3 1.25 (/home/travis/virtualenv/python3.6.3/lib/python3.6/site-packages), Requirement.parse('urllib3<1.25,>=1.21.1'), {'requests'})
```

bnmnetp:
Build is passing for 3.7 anyway...

bnmnetp:
OK @bjones1 !!  I've got travis set up to run the tests for both 2.7 and 3.7 and they are passing for both! 

I've made 3 my default environment for web2py now, and so far so good.

* Note to self, when I make the switch over in production I should delete everything in the sessions table.  Although I think we might want to move to using redis for managing sessions on academy anyway.

bnmnetp:
@bjons1 there are still some issues with Sphinx 2.0 under Python 3.7.   Lots of moving pieces!  When I allowed runestone to use Sphinx 2.0 then I got failures on server side grading of FITB questions.

bjones1:
Hmm, I'll take a look when I can. I'm at home holding down the fort today -- my son and wife have strep.

bnmnetp:
Ugh!

I’m not sure we can totally blame Sphinx as I’m not sure that has anything to do with getting a FITB answer.  It looks like it might be related to bleach as the answer returned is `<p>Correct.</p>` rather than just `Correct.`

PR #1225, opened 2019-04-27
[WIP] Adding docker-compose

vsoch:
This is a work in progress to add docker-compose (starting with @yarikoptic early changes!) I'm still figuring out the uwsgi and nginx setup, so it's not fully working yet. I'm closing up shop for today for dinner, but wanted to share what I've done so far. I'm still largely testing commands interactively, so still a bit to go :)


bnmnetp:
Thanks...  busy day today but I'll look at it tonight or first thing tomorrow.  I use uwsgi and nginx on runestone.academy so I'm sure I can help fill in the details there.

vsoch:
Sounds good! I'm working on this now again - I'm realizing that upstart doens't work well in docker so I'm looking to try something else.

vsoch:
I don't see how web2.py.py actually handles any arguments:

```
import os
import sys
from multiprocessing import freeze_support
# import gluon.import_all ##### This should be uncommented for py2exe.py

if hasattr(sys, 'frozen'):
    path = os.path.dirname(os.path.abspath(sys.executable))  # for py2exe
elif '__file__' in globals():
    path = os.path.dirname(os.path.abspath(__file__))
else:  # should never happen
    path = os.getcwd()
os.chdir(path)

sys.path = [path] + [p for p in sys.path if not p == path]

# important that this import is after the os.chdir

import gluon.widget

# Start Web2py and Web2py cron service!
if __name__ == '__main__':
    freeze_support()
    if 'COVERAGE_PROCESS_START' in os.environ:
        try:
            import coverage
            coverage.process_startup()
        except:
            print('Coverage is not available')
            pass
    gluon.widget.start(cron=True)
```

vsoch:
okay - I stepped back a bit and removed the nginx container so we could start with a simplified, working thing. The PR is updated to have a postgres and runestone container, and nginx is installed within the runestone container (no special deployment like gunicorm or uwsgi - we are just using the commands that were used previously). I'm hoping that you have enough experience with web2py to use these containers to test out deployment via another method. I seem to always have trouble figuring out a new framework with nginx + gunicorn/upstart some other nonsense :P

bnmnetp:
Sounds good.  Simple is good.   And thanks!  I'm really glad I took the 30 seconds to answer your skulpt question!

I've just been trying to get something up and running but have run into problems already.  Mostly I'm sure my ignorance about Docker. 

I think I already shot myself in the foot because I added your branch to my already locally running clone of Runestone.  This means I had a fully populated databases folder which gets the rsmanage initdb all angry at me.  I'll try tomorrow from a clean clone and see how that goes.

I'll try to figure out how to delete the containers I built today so I can restart from scratch.

vsoch:
Bringing up the containers for the first time will indeed run the database init, and look for instructors / students in the repo folder (note that it's bound to the container as code).

Definitely start with a fresh clone, and before that (from your previous repo) you will want do remove the old containers:

```bash
docker-compose stop
docker-compose rm
```
Don't forget to build the base first:

```bash
docker build -t runestone/server .
```
If you need to look at logs for a container:

```
docker-compose logs uwsgi
docker-compose logs db
```
Let me know if you have questions! It should work with a fresh clone, so let me know if I missed adding a file, etc.

vsoch:
Oh, and if you want to be able to Control+C (and see the containers all starting up) just do docker-compose without the d (which means detached):

```
docker-compose up
```

bnmnetp:
This is really nice.

I had to make a couple changes to the Dockerfile.  The critical one is to remove python-sphinx from the list of stuff that apt-get installs.  It wants to install Sphinx 1.4.9, which is ancient.

Questions:

* `restart: always` .  I've seen/read a little about this and I'm lead to believe that this would be very nice when running in development mode.  ??  Right now it seems that the workflow must be to `docker-compose stop` ; `docker build -t runestone/server .` ; `docker-compose up -d`

* Why `debian:stretch-backports`  My understanding is that debian is super conservative and generally way behind recent releases?  Stretch-backports makes me think it is even older....  Why not start with python:XX as a base??

* Is it possible to set up variations so that we could have an autoreloading development environment as well as a production environment?

Issues:

- [ ] I ran into an age old problem with the session table.  web2py throws an error that the table is a duplicate.  Need to figure out how to fix this once and for all.... likely by using redis as a much better way of dealing with session information.  -- this causes all restarts after the first one to fail.  Workaround is to set WEB2PY_MIGRATE=No in the entrypoint.sh

- [ ] Need to provide a default route file so that runestone is the default app.

- [ ] All the database migration information appears to disappear after a change and restart... I think this is the cause of  my first issue....  This is tricky as we want to preserve the database, but then we also need to preserve this migration information...  We will need to do bug fixes on both book content and server updates without starting from scratch on the database.

vsoch:
@bnmnetp in case it was leading to issues, I removed the bind to /code. This is a nice deployment strategy if you want / plan to make changes to the code base from the host, and the code is served from the place bound in the container (e.g., /code). In this case, we wind up moving files into the web2py directory, so it didn't do us much to bind, so I removed it. If we add something like it again, it would be more like:

```
     volumes:
      - .:/srv/web2py/applications/runestone
```

vsoch:
> restart: always . I've seen/read a little about this and I'm lead to believe that this would be very nice when running in development mode. ?? Right now it seems that the workflow must be to docker-compose stop ; docker build -t runestone/server . ; docker-compose up -d

Almost - to update code from the host (given the bind I've shown above) you just need to do `docker-compose restart`. You would only need to rebuild the container given some change to the Dockerfile, but you could also just shell inside and install / remove whatever needed to be (and rebuild later).

> Why debian:stretch-backports My understanding is that debian is super conservative and generally way behind recent releases? Stretch-backports makes me think it is even older.... Why not start with python:XX as a base??

You need to ask @yarikoptic about this - I started with his base and wouldn't have chosen this. He works on neurobedian stuff, hence this base and the "eatmydata" install.

> Is it possible to set up variations so that we could have an autoreloading development environment as well as a production environment?

The current setup would work to reload with docker-compose restart. The "restart" tag for the container doesn't mean restart (when there are changes to code) - it means restart if the container doesn't start successfully (an error). If / when you play around with trying different web server setups (e.g., gunicorn, uwsgi) you'll see this quite a bit :) It's why I added the (commented out) command to just send something to  /dev/null - sometimes you need the container to start (and work) so you can shell inside and poke around. If the container starts and it doesn't work, it will restart indefinitely and you can't shell in. The only case when restart is useful is if one container needs to wait on another for some reason.

bnmnetp:
So, adding something like your volumes: stuff from above would preserve `applications/runestone/databases/*` -- also `applications/runestone/static/*` which is where the books that are in the books folder  go to be served when they are deployed.

vsoch:
> So, adding something like your volumes: stuff from above would preserve applications/runestone/databases/* -- also applications/runestone/static/* which is where the books that are in the books folder go to be served when they are deployed.

Yes, if you added the bind to either the top of runestone, or those locations directly, you could bind them on the host (and then add additional books or update the runestone code) easily without needing to rebuild the container. You would need to restart, but you could do uwsgi selectively:

```
docker-compose restart uwsgi
```

> I had to make a couple changes to the Dockerfile. The critical one is to remove python-sphinx from the list of stuff that apt-get installs. It wants to install Sphinx 1.4.9, which is ancient.

Forgot to comment on this! I think @yarikoptic possibly intended to install the base and then update the version later? But if it's too old, you are correct it doesn't make sense to install at all.

>  I ran into an age old problem with the session table. web2py throws an error that the table is a duplicate. 

I'm not sure why you ran into this, because the db container should be new / fresh. Is there some way you were changing the setup to be a different database than the (unitialized one) in the docker container? I haven't run into this issue, so if you can give me enough to reproduce, I'd appreciate that.

> Need to figure out how to fix this once and for all.... likely by using redis as a much better way of dealing with session information. -- this causes all restarts after the first one to fail. Workaround is to set WEB2PY_MIGRATE=No in the entrypoint.sh

okay! Yeah I use redis for django + celery and it's good :)

>  Need to provide a default route file so that runestone is the default app.

+1 as well, I'd suggest handling these details after this original PR is good? Otherwise it gets a little large and complicated.

>  All the database migration information appears to disappear after a change and restart... I think this is the cause of my first issue.... This is tricky as we want to preserve the database, but then we also need to preserve this migration information... We will need to do bug fixes on both book content and server updates without starting from scratch on the database.

This actually isn't too hard - you will want to add a volume to the db (postgres) container. The restart shouldn't actually erase / undo the database info - I used @yarikoptic strategy and added a $stamp variable to indicate it was initialized, I didn't test if that persisted. The database shouldn't go away with a restart (restarts are common) or even a stop. But it would go away if you remove and then bring up the containers again.

bnmnetp:
Well, I don't want to leave @yarikoptic out in the cold.... But,  I would like to get a good solid installation that can help new devs get up and running quickly as well as having a pathway for better production deployment for researchers that want to run their own environments and  to a more scalable environment for runestone.academy.

bnmnetp:
I think I ran into a couple of the issues I outlined because I was stopping and rebuilding.  For example I forgot to clone a book into the books folder on my first build and start.    Then I realized that my default port and your default port were different, so I had to modify the pavement.py file for the book I was testing with and do another rebuild and restart.  So, yeah I definitely brought these on myself, but it is exactly the kind of mistake that others would make as well.

vsoch:
@bnmnetp I haven't tested adding books, and I think it would be good to have a solid set of instructions for adding them (either on init, or after running with a restart). You wouldn't need to rebuild the container to add books given the bind. Would you like me to update the PR with the bind? And could you share your changes to the Dockerfile so I can update those as well?

I'd be inclined to use a different base, but I'd like to hear @yarikoptic comments first. I stumbled on him here, and it's funny because we go way back with Singularity and neuro stuffs. Nobody gets left out in the cold on my watch!

bnmnetp:
For sure!  Would love to hear from @yarikoptic and I'm sure we can all come to an amicable agreement.

From my perspective as project lead there are at least three constituencies that I would like to support and be happy.

1.  Developers who want to get a runestone server up and running quickly, who can help add new features, bug fixes etc.

2.  Instructors like @yarikoptic and others at  institutions with the expertise and budget to run their own server for their own classes.

3.  A large and growing installation like https://runestone.academy where the scale is orders of magnitude different than 2 and 1.

Number 3 is probably the least important because that is just one installation, and as I get smarter about docker I can learn to tweak things to support that, and for the most part right now I'm doing fine running that on a digital ocean droplet.

Meanwhile I changed the docker-compose file so that the volumes: command looks like this:

```yaml
  runestone:
    image: runestone/server
    #command: tail -F peanutbutter
    volumes:
      - .:/srv/web2py/applications/runestone
    links:
      - db
```
But after rebuilding everything I now get:

```
  File "/usr/lib/python2.7/dist-packages/pkg_resources/__init__.py", line 849, in resolve
    raise DistributionNotFound(req, requirers)
pkg_resources.DistributionNotFound: The 'rsmanage' distribution was not found and is required by the application
```
This is installed as part of the server requirements.txt file with `-e rsmanage/`

So something about my change of the volumes thing is causing trouble.

vsoch:
> Number 3 is probably the least important because that is just one installation, and as I get smarter about docker I can learn to tweak things to support that, and for the most part right now I'm doing fine running that on a digital ocean droplet.

This is good insight - I've deployed (long running) services for users, and it's challenging to maintain (financially and resource wise, and just takes a lot of time!) I've found it a good strategy to open source some derivative of the server so that others can roll their own, and then I don't need to worry about managing :)

I just made all your changes, updated the docs, and reproduced your error. I'll debug and push a fix if/when I figure it out.

vsoch:
Can you tell me how the token works, in the private folder? This is generated in the runestone folder in the container (and would be over-written)

```
mkdir -p private && \
    echo "sha512:16492eda-ba33-48d4-8748-98d9bbdf8d33" > private/auth.key && \
```
What we might want to start doing is just binding subfolders within runestone with content that we want to change - I'm testing that now.

bnmnetp:
auth.key contains the HMAC key needed to crypt passwords in web2py.  If this changes then existing logins break.

vsoch:
Okay this is pretty interesting - the binary is there:

```bash
$ /usr/local/bin/rsmanage --init
```
The actually install is within runestone, and the difference (sorry for the shoddy image) is that one has an egg created (and the one from the host doesn't)

image

So we can't bind that directory, I think.

vsoch:
that image is terrible, lol. Let me see if I can just bind select directories in runestone. What is likely to change during development?

bnmnetp:
Things that will change in development:

* controllers/
* models/
* views
* static
* books
* rsmanage -- This could conceivably be moved to its own repo/package but is tightly coupled with the server.

Things that web2py needs to be persistent for a deploy

* databases -- The files in this folder are used for database migrations.  If they go missing then migrations fail and web2py crashes.
* private/auth.key . -- needed for passwords

vsoch:
okay, I've just pushed some changes to the docker-compose, and please see the (updated) docs in the docker/README.md to describe how to interact with the containers. I've tested that the docker-compose up -d brings up the application without errors, and you can further test adding books, restarting, etc.

vsoch:
Hmm, rsmanage can't be installed in the container in the Dockerfile and then also bound from the host - we can't have our cake and eat it too :) What would make this work is to remove it's install from the Dockerfile and have it done during the entrypoint (when it's bound). Would you like to try this? I would suggest leaving the installs in both places so it will work with or without the bind.

For a deployment if it's absolutely essential to have these things bound to the host, then what we would need to do is again move the steps to be redone in the entrypoint.sh.

vsoch:
For the key - wouldn't it be better to generate it on the fly (and it could be saved to the host) given the bind is there, and then it wouldn't be the same key used that is shared on GitHub (and thus a security issue?)

vsoch:
Also just  looking at your pip requirements.txt file - here is why the install is done locally:

```
-e rsmanage/
```
So let me try just doing that in the entrypoint (and making the install on the host bind)

bnmnetp:
1. - rsmanage:   In entrypoint.sh I can add a pip command to install it from there.  No problem.

2. - Now that I think about it, for development purposes we probably don't need to worry about it.  The place where it was really important is for runestone.academy so that all the logins don't break.   More recent versions of web2py don't rely on this either.  So we should not make a big issue of it.

vsoch:
okay :) Actually I think I fixed both, let me quickly test and I'll push the fixes.

bnmnetp:
cool.   could you go ahead and remove:
python-sphinx, python-requests, python-sqlalchemy, python-six from the apt install?  All of the python-* packages (except python-dev) are actually handled better by pip and the requirements.txt file.

vsoch:
Are you sure about numpy? It takes quite a long time to compile that.

bnmnetp:
Good point, we can leave numpy.

vsoch:
okay the python updates seem to work, and I've hit an issue with migrations:

```
 It appears you already have database migration information do you want to proceed? [y/N]
```
Is there a way to automatically designate yes, and if not, how would you like to proceed? Now that we have the folders bound, the migrations are saved on the host. However, if the user wants to start fresh, these would need to be removed. if the user wants to start fresh and use the migrations, then some special command (that I'm not privy to) would need to be used. Let me know what you'd like to do.

vsoch:
okay, I've updated the entrypoint to give a meaningful message if the databases folder exists, since we can't re init (and it won't work if we don't). The extra python deps are also removed. @yarikoptic when you get a chance, would be good to hear your feedback on eatmydata and the base image.

bnmnetp:
OK, this is really great.

I added a new book and did the restart and there it is, no problem.

I did run into the container not restarting because postgres was still starting up.  There must be a good way to handle this...

bnmnetp:
I can definitely add a `--force` flag to the initdb command.  But it seems like it might not be necessary now?   It would be more dangerous yet, but I could have `--force` remove the databases folder too.

vsoch:
docker-compose tends to be a little messy - we are after all bringing up separate containers (and compose is just helping to orchestrate networking, etc.). So if you issue a command to restart and it sort of lags (and then doesn't sync with the other container) the only thing to do would be to wait for postgres, or do a clean stop and then restart of all containers.

vsoch:
I would suggest not using force - if the user didn't want to delete the folders this would be problematic. The question I have is what would be the action to take if the user wanted to bring up a new set of containers with an already existing database? is this not possible?

bnmnetp:
To bring up new containers with an existing database then the thing to do is to set the WEB2PY_MIGRATE environment variable to "Fake" This causes web2py to rebuild the databases/* files based on the existing database. 

To make this easy, I should add a `--fake` option to the `rsmanage migrate` command.

bnmnetp:
So, using the `restart: always ` option is not a good strategy for handling the potential timing problem.

If not that then I could make a little enhancement to the entrypoint.sh that loops until the database is up and ready with a short 1 second time delay between tries.

bnmnetp:
Stack Overflow suggests something like:

```
command: bash -c 'while !</dev/tcp/db/5432; do sleep 1; done; npm start'
```

That doesn't seem perfect for our situation, but something like this in the entrypoint.sh was what I was thinking about:

```
RETRIES=5

until psql -h $PG_HOST -U $PG_USER -d $PG_DATABASE -c "select 1" > /dev/null 2>&1 || [ $RETRIES -eq 0 ]; do
  echo "Waiting for postgres server, $((RETRIES--)) remaining attempts..."
  sleep 1
done
```

vsoch:
+2 to add the fake option - we can leave as is for now, and when the option is added via PR, the PR can also update the command to use it (and not poop out and fail).

Have you tried restart:always? It's not currently in the file, and I don't see why it would be problematic. If you want to test that solution and form a preference for one or the other, I'd be happy to add to the PR (or you can add later if you want to work/think about it more).

vsoch:
I guess people don't normally do +2, I meant you had already done a +1, and that makes me 2. Maybe I'm doing too much programming today :)

bnmnetp:
I added the `restart: always` option and it all seems to work fine.

I'm going to go with this as my new workflow and see what kinds of issues I run into.

I'd like to hear from @yarikoptic but I think I'd like to accept this PR (with the restart: always) added.  We can always continue to refine the implementation as we get more experience using the containers in practice.

vsoch:
All set! I updated the docs too. And actually, with restart to true, we don't need to stop, remove, and bring up containers at all! The user can simply remove the databases folder and the container will restart on its own and bring up the application successfully.

bnmnetp:
Thanks @vsoch and @yarikoptic !

I'm sure this is not the last word in Dockerizing Runestone, but it is a great start, and I wanted to get it merged so that I can make it part of my daily workflow.

vsoch:
Awesome! Let me know if/when you need/want some help. I can quickly share my use case (and desire) for some of the interactive examples you've shown me - I want an easy way to have a GitHub repo where users / admins of HPC clusters can write question / answers that are then rendered automatically into a web interface (via GitHub pages) to quiz people. For example, it might be something along the lines of "How would you pull a docker hub container called "vanessa/salad" into a Singularity container?" and then the user would type "singularity pull docker://vanessa/salad"

I looked through the current examples (there is a fill in the blank) and that would work, but what wouldn't work is to have answers where (for example) order of arguments isn't important.

With a repo like this, and your "sphinx for interactive learning" we could have repos organized by topics, syndicated at a common root page for the user to start at (I'd probably make a page [here](https://vsoch.github.io/lessons/)), and within each repo a simple markdown file (or rst / similar) that is easy to add a question to.  In terms of saving state, since we are serving statically we don't have the benefit of a server (and user accounts) however we could use a local browser database (localStorage) to keep track of question progress, and reveal some badge / link back when the quiz is all completed.

As long as it's easy to add a line for a new question, this would be very easy for AskCI to contribute to. And as long as the templating is beautiful, and sort of fun, people would be incentivized to do it to get familiar with some HPC technology.

bnmnetp:
Building static standalone pages with runestone has always been a goal, and I'm glad to hear your use case as there is always pressure to move to dynamic pages.  This  reminds me that there are still good reasons to keep the ability to generate static pages.

localStorage is already used to remember answers to many, if not all, kinds of questions, so you are not far away.

Once you start a project, you can customize the template as much as you want...  But fair warning that the CSS has uhmmm evolved we'll say over 8 years starting with a professor and some students who knew very little CSS when they started this journey.

fill-in-the-blank questions support regular expressions for matching answers.  I don't suppose there would be any reason not to somehow enable an arbitrary javascript function to use in grading things where order of arguments is not important.  In fact I could imagine developing a little library of generic functions that would be easier and more powerful than regexes.

yarikoptic:
Awesome job @vsoch (well, as always!)

vsoch:
Thanks @yarikoptic ! You lit the way with the eatmydata monster :)

Dockerfile
```diff
@@ -0,0 +1,82 @@
...
+#  sphinxcontrib-paverutils -- N/A
+#  sphinx -- we need stretch-backports
+#  pytz ... ?
+# A few missing ones
+#  rsync is needed when deploying a built book
+#  vim - just for pleasure of being able to do any changes right within
+#  wget - just in case
+RUN apt-get update && \
+    apt-get install -y eatmydata && \
+    eatmydata apt-get update && echo "count 1" && \
```
yarikoptic: BTW, that "count 1" could probably be stripped -- just used it whenever wanted to force redoing this particular layer I guess

docker/entrypoint.sh
```diff
@@ -0,0 +1,88 @@
...
+    echo "I: $@"
+}
+
+# This file will exist if we've initialized postgres
+stamp=/var/lib/postgresql/9.6/main/initialized.stamp
+
+# Ensure the user starting the container has provided a password
+if [ -z "$POSTGRES_PASSWORD" ]
+then
+    echo "Please export \${POSTGRES_PASSWORD}"
```
yarikoptic: I wonder if there could be a better way which would avoid leaking any sensitive information through the environment.
I did check that skulpt has no env vars exposed, but I guess there could be other ways to see/get to it. so ideally should be via some config file

yarikoptic:
> eatmydata monster :)

at least that one doesn't eat avocados ;)  it does notably speeds up installation, so I nearly always use it when I can.  In neurodebian images I even made it to be used for any `apt-get` invocation

```shell
$> docker run -it --rm --entrypoint bash neurodebian
root@5bcf898c157c:/# ls -l `which apt-get`
lrwxrwxrwx 1 root staff 18 Mar 27 00:01 /usr/local/bin/apt-get -> /usr/bin/eatmydata
```
FWIW, if you use neurodebian images, you could make use of our experimental [nd_freeze](https://github.com/neurodebian/neurodebian/blob/master/tools/nd_freeze) which would replace APT files with the ones pointing to snapshots, so you will get guaranteed reproducible images (for the date you provide to `nd_freeze` call ;) ).  But I think it is not desired here since installing most recent security updates etc is of direct relevance

Dockerfile
vsoch: Ah okay, I figured you were channeling your inner Count!

1... 2.... 3.... hahahaaaaa !

docker/entrypoint.sh
vsoch: You are suggesting starting the database without having those variables exported? The postgres image is a fairly common "bread and butter" image, and providing the envars (in some format) is what is commonly done. We can assume if the server is compromised with the environment variables and the application, you've already lost, for the most part.

The only thing I've seen recently for Docker that allows secrets (for builds) is the --secrets flag, I had some good discussion [here](https://www.reddit.com/r/programming/comments/bcumt6/secure_secrets_in_docker_builds/) (warning, that's reddit, lol) about how it was different from a build arg. That said, I think the point of focus was removing any trace of the variable in the Docker history, or bash history.

Do you have any ideas or suggestions for how this might be done @yarikoptic, or are you thinking out loud?

vsoch:
lol I know I'm a monster :P :avocado:

docker/entrypoint.sh
yarikoptic: for postgres container it is probably ok, since no child, client-visible process will operate within postgres DB process. But here it is the entry point for the web service, and password is also below within cmdline of `python web2py.py ` invocation.  So if somehow someone gets ability to list environment variables of the process, or list of running processes -- password would leak.  How critical and possible is that -- not sure ;-)   so indeed somewhat thinking out loud with extra paranoia in the voice

vsoch: Good points @yarikoptic ! I'm not a security expert but I can see this concern. So you are saying that if there is some (web entrypoint) in the application that isn't quoted correctly, and (given python) someone could hypothetically pipe an `import os; print(os.environ.values())` into it, we would be toast.

yarikoptic: Something like that or some other way to inject some "read only" code snippet into that running webapp process.

PR #1229, opened 2019-04-30
Tweaks to get unit tests to run under Docker

bnmnetp:


bnmnetp:
@vsoch, @yarikoptic while testing this I did run across one bad loop situation:  If either addinstructor or inituser fails for some reason, like blank lines at the end of the CSV file then we eventually get stuck in a loop. because the $stamp file does not get created.

The more I think about it, I think adding users and / or instructors is not really an initialization thing.  Users and/or instructors can be added at any time at any restart if the .csv file is newer than the last time users or instructors were added.

I would also really like to get this Docker image using Python 3.  @yarikoptic what can you say about starting with `debian:stretch-backports ` ?  Would a `python:3.x` starting point work equally as well for your purposes?

vsoch:
I’d be happy to help with these things tomorrow! We can definitely move it outside of the loop, and skip empty lines. Do you have testing files that reproduce the issues you are running into?

bnmnetp:
Just documenting my thinking here, we have four scenarios

1. no database, no databases folder -- all is good run  `rsmanage initdb`
2. no database, but databases folder -- should do `rsmanage --reset` to remove the databases folder and then initialize
3. database, but no databases folder -- could try a `rsmanage migrate --fake` to create databases folder.  Or could  could do `rsmanage initdb --reset` to drop and recreate database and then initialize
4. database and databases folder -- no initialization should be needed

Case 3 is the most tricky as I am really uncomfortable just blowing away a database, so it seems that the fake migration option is the best.  If things don't work out after a fake migration then its up to the developer to decide what to do.

yarikoptic:
> I would also really like to get this Docker image using Python 3. @yarikoptic what can you say about starting with `debian:stretch-backports ` ? Would a `python:3.x` starting point work equally as well for your purposes?

I am ok with anything ATM - done teaching for a while ;)  FWIW there is python3 ins stretch-backports, but given that buster is soon to be released, you could even just switch to base your image on it instead.  Is `python:3.x` supposed to be a docker image -- can't find it (thought to check what it is based on).  As a Debianista I would prefer a Debian based image.

bnmnetp:
python is a docker image there are tags for various version of python.  various versions are supported as tags.

```
python:
This is the defacto image. If you are unsure about what your needs are, you probably want to use this one. It is designed to be used both as a throw away container (mount your source code and start the container to start your app), as well as the base to build other images off of.

This tag is based off of buildpack-deps. buildpack-deps is designed for the average user of Docker who has many images on their system. It, by design, has a large number of extremely common Debian packages. This reduces the number of packages that images that derive from it need to install, thus reducing the overall size of all images on your system.

Some of these tags may have names like jessie or stretch in them. These are the suite code names for releases of Debian and indicate which release the image is based on.
```

bnmnetp:
@vsoch I've added a flag: `rsmanage env --checkdb` will return 0,1,2,3 corresponding to cases 1,2,3,4 above.

You will see that I moved some environment variables to the Dockerfile so that they are available when I shell in to the container to do things like run unit tests.  Should POSTGRES_PASSWORD be made an ARG ??  I assume that if they are in the Dockerfile I can remove them from the entrypoint.sh ?

vsoch:
A few quick points to address (I can work on this in a bit but not immediately!)

 - The python 3.7 base is based on [debian stretch](https://github.com/docker-library/python/blob/master/3.7/stretch/Dockerfile). If some of the libraries require earlier versions of 3.* I would assume having the same base images. If you want to get tiny you could consider alpine, but you would need to find the corresponding packages (not impossible, but a redo generally).
 - @bnmnetp would you like me to take a stab at changing some of the init sequence or do you just want to have conversation on details here (and you are working on this pr?)
 - `POSTGRES_PASSWORD` as it was originally set up is optimized for development - coming from the docker-compose file. If you have it there, you don't need it as an ARG. I could be wrong, but I think the docker-compose should supply the envars wouldn't having them as args. I would keep the check in the entrypoint.sh to ensure that it's defined, otherwise you would hit an error. There could be cases of a user testing some other way to provide the variables and not being successful (and thus this catch doesn't hurt to have).

What action items do you want some help with?

bnmnetp:
Thanks, I'm using this PR as a chance to get more familiar with how docker and docker-compose work.  So I'm grateful to have you as an advisor on this one.

vsoch:
okey doke :) As they say in Warcraft 2... Ready to Serve! :) And that's a pun because you are also serving a web server :o)

yarikoptic:
<details>
<summary>ah, sorry, I just took `python:3.x` too literally ;)</summary>

```shell
$> docker run -it --rm python:3.x
Unable to find image 'python:3.x' locally
docker: Error response from daemon: manifest for python:3.x not found.
See 'docker run --help'.

$> docker run -it --rm python:3.7
Unable to find image 'python:3.7' locally
3.7: Pulling from library/python
e79bb959ec00: Already exists 
d4b7902036fe: Downloading  4.357MB/10.78MB
1b2a72d4e030: Download complete 
d54db43011fd: Downloading  5.594MB/50.07MB
69d473365bb3: Downloading  535.2kB/215MB
...
```
</details>

vsoch:
@yarikoptic how do you do the little dropdown arrow in markdown? I haven't seen that yet (it's super cool!)  It's definitely useful for all those debugging outputs that one might copy paste into an issue!

vsoch:
@bnmnetp good call to move the environment variables (and now I see what @yarikoptic was referring to). For a production deployment, these would either be in a (non public) compose file or from a separate .env file on the host, read in by compose.

yarikoptic:
> @yarikoptic how do you do the little dropdown arrow in markdown? I haven't seen that yet (it's super cool!) It's definitely useful for all those debugging outputs that one might copy paste into an issue!

as pure HTML with markdown embeding

    <details>
    <summary>title ...</summary>
    
    ```shell
    that empty line before this block is important somehow
    ```
    </details>

vsoch:
<details>
<summary>What is my purpose?</summary>

```shell
I serve hidden text dropdowns! :D 
```
</details>

vsoch:
I love it! Can I share with others?

yarikoptic:
> I love it! Can I share with others?

how could I say "No"? ;)

vsoch:
I mean, you always could. :)

bnmnetp:
@vsoch , @yarikoptic could you take a look at entrypoint.sh 

I made a lot of changes, but I think the startup is more robust now... Especially for people who might be recreating Docker containers.  And in general I moved some things to get the server up quicker on a restart.  If books are already built then waiting for them to rebuild before you can get access to the server is frustrating.

Thanks

Brad

vsoch:
Sure, testing now.

vsoch:
Okay quick feedback - the docker history looks good in that I don't see secrets, etc.
<details>

```shell
$ docker history runestone/server
IMAGE               CREATED             CREATED BY                                      SIZE                COMMENT
fbfd515252e0        6 minutes ago       /bin/sh -c #(nop)  CMD ["/bin/sh" "-c" "/bin…   0B                  
5ce10a220545        6 minutes ago       /bin/sh -c #(nop) COPY file:6a9549e761c1aa0b…   3.54kB              
1ea5936c185b        6 minutes ago       /bin/sh -c #(nop) WORKDIR /srv/web2py           0B                  
960f69b210a2        6 minutes ago       |3 DEBIAN_FRONTEND=noninteractive WEB2PY_APP…   247MB               
ea2efbeaf755        7 minutes ago       /bin/sh -c #(nop) WORKDIR /srv/web2py/applic…   0B                  
3e1634f84efb        7 minutes ago       /bin/sh -c #(nop) ADD dir:2b9aa10b04214d5ad1…   291MB               
8fff787be144        7 minutes ago       |3 DEBIAN_FRONTEND=noninteractive WEB2PY_APP…   0B                  
c87eafba496c        7 minutes ago       |3 DEBIAN_FRONTEND=noninteractive WEB2PY_APP…   96MB                
f486af62423b        7 minutes ago       |3 DEBIAN_FRONTEND=noninteractive WEB2PY_APP…   327kB               
5be334f09b24        7 minutes ago       |3 DEBIAN_FRONTEND=noninteractive WEB2PY_APP…   563MB               
74f5e715d5d8        8 minutes ago       /bin/sh -c #(nop)  ARG DEBIAN_FRONTEND=nonin…   0B                  
92ac3a9c11b3        8 minutes ago       /bin/sh -c #(nop)  EXPOSE 8080                  0B                  
29582747f3a8        8 minutes ago       /bin/sh -c #(nop)  ENV DBHOST=db                0B                  
4b11b9bfa165        3 days ago          /bin/sh -c #(nop)  ENV WEB2PY_VERSION=2.18.4    0B                  
4a393915cfb2        3 days ago          /bin/sh -c #(nop)  ENV BOOKS_PATH=/srv/web2p…   0B                  
4bd9c8f5194b        3 days ago          /bin/sh -c #(nop)  ENV RUNESTONE_PATH=/srv/w…   0B                  
3c2f4269d354        3 days ago          /bin/sh -c #(nop)  ENV WEB2PY_PATH=/srv/web2…   0B                  
6a10bf57d3e2        3 days ago          /bin/sh -c #(nop)  ARG WEB2PY_PORT=8080         0B                  
207c63b4807d        3 days ago          /bin/sh -c #(nop)  ARG WEB2PY_APPS_PATH=/srv…   0B                  
b2aa60b73897        3 days ago          /bin/sh -c #(nop)  ARG WEB2PY_PATH=/srv/web2…   0B                  
499c91b427ac        3 days ago          /bin/sh -c #(nop)  LABEL authors=@bnmnetp,@v…   0B                  
f20b63681cba        5 weeks ago         /bin/sh -c echo 'deb http://deb.debian.org/d…   56B                 
<missing>           5 weeks ago         /bin/sh -c #(nop)  CMD ["bash"]                 0B                  
<missing>           5 weeks ago         /bin/sh -c #(nop) ADD file:843b8a2a9df1a0730…   101MB      
```
</details>

In terms of the Dockerfile, I'm wondering if the DB host should be an ARG, that way if someone wants to use the container (but not the docker-compose postgres) then they can build and include as a build arg. So instead of this:

```
ENV DBHOST=db
```
you would do

```
ARG DBHOST=db
```
And then when someone is building the container, they would do:

```bash
$ docker build --build-arg DBHOST="${THE_BIG_DB}" -t runestone/server .
```
I don't see moving the various envars out of the entrypoint, so I assume that's in a different PR. Oh hmm, is there a reason you have it both in the entrypoint and the docker-compose? The docker-compose would need to test if you can define variables based on other variables (e.g., if the password, provided as it's own variable, feeds into the database url).

Can you add a comment above:

```bash
if [ ! -f "${RUNESTONE_PATH}/models/1.py" ]; then
        touch "${RUNESTONE_PATH}/models/1.py"
    fi
```
that explains the check, and what it's doing?

The first comment for students should say instructors.

I like the switch that determines different states, this is a much better solution than a binary choice of "wipe everything and have nothing" or "won't work at all."

Why do you build the books after starting the server (intuitively I'd do it the other way around).

Why did you decide to add the restart loop, in case someone doesn't use compose?

Probably want to be cautious about having the password / dburl in the logs:

<detail>

```shell
runestone_1_8e2941527864 | Database migration folder exists
runestone_1_8e2941527864 | WEB2PY_CONFIG is production
runestone_1_8e2941527864 | The database URL is configured as postgresql://runestone:runestone@db/runestone
runestone_1_8e2941527864 | DBNAME is runestone
runestone_1_8e2941527864 | Exiting with result of 3
runestone_1_8e2941527864 | I: Got result of 3
runestone_1_8e2941527864 | I: All is good, no initialization needed
runestone_1_8e2941527864 | I: Starting the server
runestone_1_8e2941527864 | I: Building & Deploying books
runestone_1_8e2941527864 | I: Starting the scheduler
runestone_1_8e2941527864 | web2py Web Framework
runestone_1_8e2941527864 | Created by Massimo Di Pierro, Copyright 2007-2019
runestone_1_8e2941527864 | Version 2.18.5-stable+timestamp.2019.04.07.21.13.59
runestone_1_8e2941527864 | Database drivers available: sqlite3, psycopg2, pymysql, imaplib
runestone_1_8e2941527864 | 
runestone_1_8e2941527864 | please visit:
runestone_1_8e2941527864 | 	http://127.0.0.1:8080/
runestone_1_8e2941527864 | use "kill -SIGTERM 18" to shutdown the web2py server
runestone_1_8e2941527864 | 
runestone_1_8e2941527864 | 
runestone_1_8e2941527864 | starting scheduler for "runestone"...
runestone_1_8e2941527864 | Currently running 1 scheduler processes
```
</detail>

Other than that, I was able to bring it up, bring it down (with a created database) and then up again without issue, and I got a warning for a blank password. Just wondering, where is the password being found to be blank (and then shown to be runestone in the print above?)

bnmnetp:
Thanks for the feedback.  Heading out to dinner with friends in a moment.  Will look at it first thing tomorrow.

The reason I moved the (re)build of the books until after was that it is frustrating, when in dev mode, to wait for the rebuild to complete before starting the server.  The rebuild can take several minutes.

The restart loop can be re-thought now that we have handled the database initialization better.  Before these changes,  if the database didn't come up in time on a restart we got stuck in a fruitless loop.

vsoch:
Okay, that makes sense!

And a fruitless loop! Oh dear. If I'm going to go into a loop, we definitely need to have froot loops.

Have a great dinner! :plate_with_cutlery:

bnmnetp:
I think I've addressed most things...  

I like having dbhost as an arg especially for the case where someone might have a managed database they want to use instead of spinning up their own.

I've hidden the DBURL when `rsmanage env --checkdb` is called.

I left the check for the database being up in the entrypoint as it just seems more trustworthy and clean than failing and restarting over and over.  We don't want to keep people waiting for a container restart when they are developing.  We can still revisit that later.

I'd like to merge this so I can spend some time and attention on a couple of other high priority PR's.  But I would really like to return to the issue of python3 soon. (next week is pretty much doomed)

I did discover a really nice script in the `web2py/scripts` folder called `setup-web2py-nginx-uwsgi-ubuntu.sh`  I know the python container is based on debian, but maybe this can still be used or at least adapted to quickly set up an environment that would be largely identical to what I run in production.

rsmanage/rsmanager.py
```diff
@@ -385,14 +385,17 @@ def env(config, checkdb):
...
-    echoEnviron(config)
+        if os.path.exists(DBSDIR):
+            dbdir = 1
+            print("Database migration folder exists")
+        else:
+            dbdir = 0
+            print("No Database Migration Folder")
+
+    if not checkdb or config.verbose:
+        echoEnviron(config)
```
vsoch: Is this going to print passwords / etc. in clear sight? You have to be careful with docker-compose / docker - all of these printed statements will be (forever, pretty much) in the logs.

vsoch:
@bnmnetp I found that script too! I didn't give it justice to test out, so we can definitely try again. Since I was unfamiliar with the application and web2py I opted to try and get the current deployment approach working in containers. We can try that next!

I'd love to help with python3 conversion - has someone started this yet?

bnmnetp:
No, so if you have the time that would be awesome.

PR #1230, opened 2019-05-01
Fix: Update tests to allow older and newer Sphinx output to pass.

bjones1:


bnmnetp:
Wow, thats an interesting change on the part of Sphinx.  I'm surprised it didn't break a whole lot of other things.

PR #1231, opened 2019-05-02
Updates for Python 3

vsoch:
The work to make (at least the superficial running) work with Python 3 was fairly simple -

 1. exporting encoding environment variables supported (see locale -a for the three) so that click didn't bug out.
 2. updating the container installs to python3-X
 3. Creating a symbolic link from python3 to 2 so "python3" isn't hard coded everywhere.
 4. There was one strange error that web2py didn't like having "runestone" at the end. The -X variable is a flag for running with the scheduler, so I'm guessing that wherever the args are ultimately parsed (the gluon thing?) has a difference between Python 2/3. It seems to work okay removing it, so I'll want your feedback on this.

I suspect that the RunestoneInteractive/RunestoneComponents might need work, along with some of the books? I wasn't sure how to properly run tests, so that's something else to do (and I'm hoping this PR kicks that off). Hopefully this is a good start!

Signed-off-by: Vanessa Sochat <vsochat@stanford.edu>

### DIFF
--- a/ChangeLog.rst
+++ b/ChangeLog.rst
@@ -1,11 +1,11 @@
 Runestone ChangeLog
 ===================
 
-3.5.x August 2015
+3.5.x August 2015 - X 2019
 -----------------
 
 * Addition of Dockerfile for container deployment
-
+* Conversion to Python 3
 
 3.1.1 September 1, 2014
 -----------------------
--- a/Dockerfile
+++ b/Dockerfile
@@ -19,6 +19,9 @@ ENV RUNESTONE_PATH=${WEB2PY_APPS_PATH}/runestone
 ENV BOOKS_PATH=${RUNESTONE_PATH}/books
 ENV WEB2PY_VERSION=2.18.4
 
+# Click needs these encodings for Python 3
+ENV LC_ALL=C.UTF-8
+ENV LANG=C.UTF-8
 
 # Expose that port on the network
 EXPOSE ${WEB2PY_PORT}
@@ -45,14 +48,13 @@ RUN apt-get update && \
         gcc \
         git \
         unzip \
-        python-pip libfreetype6-dev postgresql-common postgresql postgresql-contrib \
+        python3-pip libfreetype6-dev postgresql-common postgresql postgresql-contrib \
         libpq-dev libxml2-dev libxslt1-dev \
-        python-setuptools \
-        python-numpy \
-        python-dev \
-        python-wheel rsync wget nginx && \
+        python3-setuptools \
+        python3-numpy \
+        python3-dev \
+        python3-wheel rsync wget nginx && \
     rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/*
-
 
 # The rest could be done and ran under a regular (well, staff for installing under /usr/local) user
 RUN useradd -s /bin/bash -M -g staff --home-dir ${WEB2PY_PATH} runestone && \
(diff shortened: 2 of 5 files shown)


bnmnetp:
Thanks, I'll take a look at the scheduler. -- that is the process that handles long running tasks, like building a book for a new class.  But that will be moot after I accept #1112 .

The only task right now that uses the scheduler is to build a new static book for a course.  After #1112 You will still be able to build static books using RunestoneComponents, for the use case where you want to host them yourself --  but as far as RunestoneServer is concerned we won't have much use for scheduler.  Yay! Less things to break. 

For most installations each course will be served as dynamic pages.  We are close to the best of both worlds where you can build a book once to have to it available on your device with no internet access, OR as a dynamic page.  Or for your use case @vsoch you can build and serve from github pages.

bnmnetp:
PS. -- the RunestoneComponents have been python3 compatible for a long time.... We have been waiting for web2py to become python3 compatible...

bnmnetp:
My only issue, now that I've built the container is that it installs Python3.5 . using FROM python:3.7 only required minor tweaks (remove --system from the pip install, and rm /usr/bin/python before linking python3. Still based on Debian, and I know there are places where I've used f-strings in my code that are not compatible with 3.5.

vsoch:
Those seem totally reasonable! I'll make the edits.

vsoch:
okay, updates are made, and I (briefly) tested that the container comes up! Python version:
```
# python --version
Python 3.7.3
```

bnmnetp:
Thanks! -- just out of curiosity is there a difference between 3.7 and 3.7-stretch ??

Also, it would be very useful for the book building process to know the hostname of the host. Not the running container, but the machine I'm running `docker-compose up` from.  Is there a way to set that as an environment variable, I guess it would need to go in the docker-compose.yml file?

vsoch:
Currently, no - if you [look here](https://hub.docker.com/_/python) you can see each row of tags - each row of tags is pointing to one Dockerfile, and the exact tag that we choose will specify the granularity of our choice. For example, if you search for 3.7 on the page you will see a windows, two alpine, and the stretch (debian) base. You can then [see this reflected](https://github.com/docker-library/python/tree/65b5603fdf177c6d90df6ea6f3bd6e23839c63a4/3.7) on GitHub when you look at the Dockerfiles.

To answer your question, we choose the tag based on how much change we are willing to accept. `python:3.7` is a moving target because when they decide to update the distro to something other than stretch, that is what we will get. Something like `python:3.7-stretch` will nail down the version, but it won't limit us to the potential of releasing another stretch with 3.7 (perhaps with security updates). The only way to nail down an exact container is to use the shasum, but (for most things) that is probably overkill.

Yes, the hostname could be passed in via an environment variable or similar.

vsoch:
It looks like 3.7 tests failed - is this a travis thing? (seems to be url errors). The tests were passing before, so I want to make sure the tweaks I just made didn't lead to the failure.

bnmnetp:
Looks like a transient failure...  I see that on my own machine occasionally, so I think we need to beef up that test.

vsoch:
:green_apple: :green_book: :green_heart: :green_salad: !

bnmnetp:
Are you saying that our tests should eat more greens??

vsoch:
:rabbit2:

bnmnetp:
93 results on Amazon for "how to speak emoji"

vsoch:
I call this mostly white piece of art "bunny in a snowstorm." He does eat a lot of greens :)

PR #1232, opened 2019-05-13
config files for uwsgi + nginx

bnmnetp:
These config files are based on the config files from my setup on Digital Ocean running an Ubuntu 16.04 droplet.  This does seem to be compatible with the debian-stretch image that we use as our base.

- [x] copy docker/nginx/sites-available/runestone to /etc/nginx/sites-enabled
- [x] copy docker/uwsgi/sites/runestone.ini to /etc/uwsgi/sites/runestone.ini
- [x] copy docker/systemd/system/uwsgi.service to /etc/systemd/system/uwsgi.service
- [x] copy docker/wsgihandler.py to /srv/web2py/wsgihandler.py
- [x] enable both uwsgi and nginx to start up 

- [x] Expose the nginx service on some port probably not 80 as it is here?

On digital ocean things are owned and run as the bmiller user.  I've tried to update here to have them owned and run as www-data.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -54,7 +54,8 @@ RUN apt-get update && \
         python3-setuptools \
         python3-numpy \
         python3-dev \
-        python3-wheel rsync wget nginx && \
+        python3-wheel \
+        rsync wget nginx uwsgi uwsgi-plugin-python3 && \
     rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/*
 
 # The rest could be done and ran under a regular (well, staff for installing under /usr/local) user
@@ -77,7 +78,6 @@ RUN mkdir -p private && \
     echo "sha512:16492eda-ba33-48d4-8748-98d9bbdf8d33" > private/auth.key && \
     pip3 install -r requirements.txt && \
     pip3 install -r requirements-test.txt && \
-    pip3 install uwsgi && \
     rm -rf ${WEB2PY_PATH}/.cache/* && \
     cp ${RUNESTONE_PATH}/scripts/run_scheduler.py ${WEB2PY_PATH}/run_scheduler.py && \
     cp ${RUNESTONE_PATH}/scripts/routes.py ${WEB2PY_PATH}/routes.py
@@ -94,5 +94,7 @@ COPY docker/nginx/sites-available/runestone /etc/nginx/sites-enabled/runestone
 COPY docker/uwsgi/sites/runestone.ini /etc/uwsgi/sites/runestone.ini
 COPY docker/systemd/system/uwsgi.service /etc/systemd/system/uwsgi.service
 COPY docker/wsgihandler.py /srv/web2py/wsgihandler.py
+RUN ln -s /etc/systemd/system/uwsgi.service /etc/systemd/system/multi-user.target.wants/uwsgi.service
+RUN rm /etc/nginx/sites-enabled/default
 
 CMD /bin/bash /usr/local/sbin/entrypoint.sh
--- a/Dockerfile
+++ b/Dockerfile
@@ -48,6 +48,7 @@ RUN apt-get update && \
         gcc \
         git \
         unzip \
+        emacs-nox \
         python3-pip libfreetype6-dev postgresql-common postgresql postgresql-contrib \
         libpq-dev libxml2-dev libxslt1-dev \
         python3-setuptools \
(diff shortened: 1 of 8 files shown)


bnmnetp:
HI @vsoch - Any advice on what should be done to the docker-compose file to expose the nginx service?

Also, just to confirm, I think that all of the copying of config files can be done with COPY commands in the Dockerfile, yes?

vsoch:
This looks great! If you are using docker-compose, you can either copy the file into the container a la Dockerfile (and it will be served from there, e.g.,)

```
COPY <src> <dest>
```
or if it's a file that you intend to be changed frequently from the host, or some kind of credential that you wouldn't want laying around in a container, you can add to the docker-compose as a volume. For these config files - most look okay to just copy.

vsoch:
And if you are adding nginx (in the same container) to serve the original 8080 proxied to 80 (does your current config just deploy to localhost:80? then you would just need to change the -ports exposed to be 80. For example:

```
    ports:
      - "8080:8080"
```
to

```
    ports:
      - "80:80"
```
and then when you add https:

```
    ports:
      - "443:443"
```
For https, I would maintain two docker-compose.yml files, with the default being just port 80 (the likely use case) and then a second with https, and also the binds to your keys. For the average user that is developing, they won't need to do any additional work. For the user that wants to deploy with https (more rare) you can include instructions to change the file, generate the certificates, etc.

bnmnetp:
Thanks.

So, the other TODO item will be to update the entrypoint.sh script so that it does not start its own copy of web2py.  That is handled by nginx and the wsgihandler.py now.  But if I'm understanding correctly entrypoint.sh should not exit.

vsoch:
What is the new entrypoint, and does it handle installing the books, etc.? If the server start command is just different, you can edit entrypoint.sh to use it at the end. If the entire entrypoint.sh is now useless / deprecated, then you'd want the new entrypoint to be whatever command is now used to perform the same operations.

bnmnetp:
The current entrypoint still does a lot of useful stuff.  Building books, populating students,  initializing the database if needed, etc.  I'm still getting used to the container mindset which is very different than how things work on runestone.academy -- configure once a long time ago and let it run for months and months without stopping...

entrypoint.sh  no longer needs to start web2py, but all of the other tasks are still good.

I think that we might add something like:n

```
service uwsgi start
service nginx start
```
At the end of entrypoint.  Rather than using systemctl to enable them to startup automatically on a restart.

vsoch:
Yep that would work, did you test it? You'd want to make sure it worked okay for restarts, etc.

bnmnetp:
This is still a work in progress... Testing but not working yet.

Is it considered good practice to have everything run as root in a container?  Some things like nginx are set up to be otherwise, and I think this is causing some permissions problems for me in my testing.

vsoch:
It's definitely not best practice (e.g., you can create a user in the Dockerfile with `USER pancakes`) but it tends to be easier to have everything done as root. I would first try to do it "the better way" and create users for services. The trick is whatever user is active when you finish the build, that's going to be the user that is present when you start the container. I would get it working first with root, and then see if you can customize to be different users. There is some conversation and examples here, if it helps:

 - https://github.com/docker-library/nginx/issues/4
 - https://github.com/nginxinc/docker-nginx-unprivileged

bnmnetp:
I'm sure this is just some newbie problem but hopefully you can help.

So I added

```
COPY docker/nginx/sites-available/runestone /etc/nginx/sites-enabled/runestone
COPY docker/uwsgi/sites/runestone.ini /etc/uwsgi/sites/runestone.ini
COPY docker/systemd/system/uwsgi.service /etc/systemd/system/uwsgi.service
COPY docker/wsgihandler.py /srv/web2py/wsgihandler.py
```
To the Dockerfile -- just like the copy commands to copy entrypoint.sh

But none of the files are there!

```
Step 25/29 : COPY docker/nginx/sites-available/runestone /etc/nginx/sites-enabled/runestone
 ---> 91346596c5dd
Step 26/29 : COPY docker/uwsgi/sites/runestone.ini /etc/uwsgi/sites/runestone.ini
 ---> ff0a65085466
Step 27/29 : COPY docker/systemd/system/uwsgi.service /etc/systemd/system/uwsgi.service
 ---> ad4674204fd7
Step 28/29 : COPY docker/wsgihandler.py /srv/web2py/wsgihandler.py
 ---> d519c7c7625b
Step 29/29 : CMD /bin/bash /usr/local/sbin/entrypoint.sh
 ---> Running in b05059f3ef0b
Removing intermediate container b05059f3ef0b
 ---> 3e092293e5dc
Successfully built 3e092293e5dc
Successfully tagged runstone/server:latest
```

Looks like it was successful but ???

vsoch:
That's not logical - I think likely you are starting / stopping docker-compose and using a previously built container. Try doing:

```
docker-compose stop 
docker-compose rm
```
or just shell directly into a fresh run of the container you just built

```
docker run -it --entrypoint bash runestone/server 
```

bnmnetp:
Yeah, I'm definitely getting an old image as my changes to entrypoint are not there either.

Ugh!  It was a typo on my build command.  `runstone/server` instead of `runestone/server` . But I'm surprised that `docker-compose up ` did not complain!  I had done an rm which I thought was removing the images...

vsoch:
docker-compose... he works in mysterious ways! :)

bnmnetp:
OK, this is working for me now!  

I'm a little worried about using the master from web2py but I guess thats probably OK for an image that is meant for development more than production.

vsoch:
Awesome!

PR #1233, opened 2019-05-16
Fix for docker instructions

sahi1l:
I tried running the unit tests with `runestoneserver_runestone_1`; I believe it should be `runestone_runestone_1` instead. Also changed an "it's" to "its".


bnmnetp:
Hi Scott,

So, you need to read up on using branches in your github development.  Your personal changes and your two pull requests are all mixed together and both pull requests contain everything.

You will want to make a branch for each pull request, and make sure that you don't commit personal changes like changing the port to 81.

sahi1l:
I've added these Docker changes to my other pull request, which doesn't have the personal stuff. (Sorry, still getting the hang of this.)

bnmnetp:
No problem.  It takes a while to learn all github charms.

PR #1234, opened 2019-05-18
Added titles to the instructor's pages

sahi1l:
The instructor's admin pages all show the title "Runestone Interactive" in the tab, which is confusing if you have multiple tabs open (Assignments and Grades, for instance).  This fix changes the tab names to correspond to the particular page.  I think this is the right way to do it?

### DIFF
--- a/static/js/admin.js
+++ b/static/js/admin.js
@@ -411,14 +411,17 @@ function getRightSideGradingDiv(element, acid, studentId) {
 
 
 function updateColumn2() {
-    var select1 = document.getElementById("gradingoption1");
-    var val = select1.options[select1.selectedIndex].value;
-    var select = document.getElementById("gradingoption2");
-    var val2 = select.options[select.selectedIndex].value;
-    var select2 = document.getElementById("gradingcolumn1");
-    var column2 = document.getElementById("gradingcolumn2");
-    var selectedval = select2.options[select2.selectedIndex].value;
-    if (val == 'assignment') {
+    var SEL1 = document.getElementById("gradingoption1");
+    var VAL1 = SEL1.options[SEL1.selectedIndex].value;
+    var SEL2 = document.getElementById("gradingoption2");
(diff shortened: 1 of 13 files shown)


sahi1l:
The unit tests don't seem to like the fact that I put the line `{{response.title="Assignments"}}` above the `{{extend}}` in assignments.html (e.g.), even though it works fine and seems to be valid web2py syntax. (cf http://web2py.com/books/default/chapter/29/05/the-views).  Not sure what to do about it.

bnmnetp:
The problem is that you are trying to do this in the wrong place.  the proper place to set response.title is in the controller, not in the view.  So for assignments it is in admin.py in the function called assignments.  for grding its in admin.py in the function called grading.  

The key to unlock the code is that all the URLs follow the pattern of `/application/controller/function`

So `/runestone/admin/grading` means to load the admin.py file from controllers and call the grading function.  You can just add a line in there to set response.title and there is already a place for that value to be filled in in the master view (called layout.html)

bnmnetp:
I'm not sure why your container has a different name than mine.  For me it is definitely `runestoneserver_runestone_1`

sahi1l:
OK, everything passes.  Now, did I do this pull request right or should I go back and separate it into smaller, more thematically coherent chunks?

static/js/admin.js
```diff
@@ -409,108 +409,119 @@ function getRightSideGradingDiv(element, acid, studentId) {
 
 }
 
-
+function makeOption(text,value,disabledQ){
```
bnmnetp: These changes seem out of place and very unrelated to the purpose of your PR.  I’m not sure who all of these changes got here.

sahi1l: OK, I'll replace this with separate pull requests: is the key fewer and more substantial changes, rather than smaller and more frequent changes? Or do all y'all have like a hundred different branches, one for each fixed typo?

sahi1l:
I'll resubmit these as multiple changes, now that I've got the idea.

PR #1240, opened 2019-05-29
Add: Server-side support for literate programming question types.

bjones1:
This is the server side for https://github.com/RunestoneInteractive/RunestoneComponents/pull/817. It won't pass until that PR is merged.

### DIFF
--- a/controllers/admin.py
+++ b/controllers/admin.py
@@ -34,6 +34,7 @@ AUTOGRADE_POSSIBLE_VALUES = dict(
     poll=['interact'],
     page=['interact'],
     showeval=['interact'],
+    lp_build=ALL_AUTOGRADE_OPTIONS,
     reveal = []
 )
 
@@ -54,7 +55,8 @@ WHICH_TO_GRADE_POSSIBLE_VALUES = dict(
     poll=[],
     reveal=[],
     showeval=ALL_WHICH_OPTIONS,
-    page=ALL_WHICH_OPTIONS
+    page=ALL_WHICH_OPTIONS,
+    lp_build=ALL_WHICH_OPTIONS,
 )
 
 # create a simple index to provide a page of links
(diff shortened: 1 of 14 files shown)


bnmnetp:
I merged #817 but I'm still getting a fail on the tests.

```
Extension error:
Could not import extension CodeChat.CodeToRestSphinx (exception: No module named 'sphinx.project')
```

bnmnetp:
@bjones1 It appears taht sphinx.project is sphinx >= 2.0 ??

I upgraded to 2.0.1 and that error went away.

But I am seeing:

```
Running Sphinx v2.0.1
WARNING: while setting up extension runestone.lp: role 'docname' is already registered, it will be overridden
```

Is that intentional?  Seems like a bad idea to override an existing role, unless it is compatible but does more?

bjones1:
Oops -- I didn't think about that case. Yes, CodeChat requires Sphinx 2.0.

The warning is ignorable -- the new `docname` role is a Runestone/Sphinx-customized version of the more generic `docname` role provided by CodeChat for docutils.

bnmnetp:
Yeah, CodeChat also causes the 2.7 build/test to fail when it tries to install CodeChat

I updated the master:requirements.txt to specify sphinx>=1.8.5 which works for sphinx but CodeChat definitely breaks anything with 2.7

bnmnetp:
Previously, the only purpose of having the scheduler running was for building books.  It looks like you have some other need for that in this PR??

bjones1:
Hmm, CodeChat should only be installed for Python 3 -- in `requirements.txt`, it is:

`CodeChat; python_version != '2.7'`

I use the scheduler to run builds/simulations on the webhost.

bnmnetp:
OK, good to know.  I had disabled the scheduler in Docker thinking we were through with it!  Then when I merged your PR onto my dev branch everything started crashing due to an import problem with scheduler!

bnmnetp:
It would be great to have a small PR against the overview book that shows how this is used.

Also, if you  could add a docstring to class _LpBuildButtonDirective then the `runestone doc lp_build` command will show the documentation for it like it does for other runestone directives.

PR #1241, opened 2019-05-29
Misc fixes

bjones1:
A couple of minor tweaks. The subtle one I found in testing:

1. A user logs in, does stuff, then logs out.
2. Using the same PC and web browser, another user then hits the `hsblog` endpoint without logging in. Then, the `compareAndUpdateCookieData` function sees the two users as identical. It therefore updates the `useinfo` table with the second user's data, which actually overwrites values the first user submitted. I added an extra check to prevent overwriting a logged-in user with a logged-out user's data.

### DIFF
--- a/controllers/ajax.py
+++ b/controllers/ajax.py
@@ -32,7 +32,7 @@ EVENT_TABLE = {'mChoice':'mchoice_answers',
 
 
 def compareAndUpdateCookieData(sid):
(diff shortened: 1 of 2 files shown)


bnmnetp:
This is exactly the kind of thing that I am looking forward to eliminating by revamping the login process for dynamic pages.

bjones1:
Yes, I can't wait! This code is tough to maintain.

PR #1242, opened 2019-05-30
Add: is_testing setting.

bjones1:
As @presnick mentioned, this is the beginning of a testing mode, where feedback isn't shown to the user. So far, it only supports fitb questions.

### DIFF
--- a/models/0.py
+++ b/models/0.py
@@ -32,6 +32,9 @@ settings.server_type = "http://"
 settings.academy_mode = True
 settings.lti_only_mode = False
 settings.coursera_mode = False
+# True to enable a testing environment, where feedback on certain question types
+# is not provided to the student.
+settings.is_testing = False
 
 # Do not control this with hostnames
 config = environ.get("WEB2PY_CONFIG", "NOT SET")
--- a/controllers/ajax.py
+++ b/controllers/ajax.py
@@ -803,6 +803,8 @@ def gettop10Answers():
     course = request.vars.course
     question = request.vars.div_id
     response.headers['content-type'] = 'application/json'
+    if settings.is_testing:
+        return json.dumps([ [], [] ])
     rows = []
 
     try:
@@ -949,6 +951,9 @@ def getAssessResults():
             return ""   # server doesn't have it so we load from local storage instead
         answer = json.loads(rows.answer)
         correct = rows.correct
+        if settings.is_testing:
+            answer['resultString'] = answer['resultString'].rsplit('Test results:\n', 1)[0]
+            correct = None
         return json.dumps({
             'answer': answer,
             'timestamp': str(rows.timestamp),
--- a/modules/feedback.py
+++ b/modules/feedback.py
@@ -116,10 +116,16 @@ def fitb_feedback(answer_json, feedback):
                 displayFeed.append(feedback_for_blank[-1]['feedback'])
 
     # Return grading results to the client for a non-test scenario.
-    res = dict(
-        correct=correct,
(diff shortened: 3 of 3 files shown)


models/0.py
```diff
@@ -32,6 +32,9 @@
 settings.academy_mode = True
 settings.lti_only_mode = False
 settings.coursera_mode = False
+# True to enable a testing environment, where feedback on certain question types
+# is not provided to the student.
+settings.is_testing = False
```
bnmnetp: settings is a site-wide variable that will effect all books currently being served.  Is this is a good strategy in a context where we are going to be serving many books simultaneously...  This seems like something that should be set and included as a part of the book rendering process, for a particular page...

bjones1: That's a good point. Actually, it opens up a whole set of questions:

1. How should feedback hiding work?
    1. Statically -- feedback is either always hidden or always shown until the student submits another answer or the page is refreshed.
    2. Dynamically -- the instructor can show/hide feedback, and students will see immediate updates. Doing this means that adding server push ability, a level of complexity I want to avoid.
2. What granularity should we support?
    1. Site-wide -- the current approach. I currently have a `runestone` application with all answers shown, and another `exam_runestone` application with all answers hidden. Both applications use the same course name, which means there's no extra steps for me to duplicate enrollment between a normal class and a class used only for exams.
    2. Book-wide -- That seems reasonable. However, it breaks my model above -- now, I need to create another course with the same enrollment as the non-exam course.
    3. Chapter/subchapter/page-wide -- sounds nice. If there was an exam "chapter", I could make all problems there hide feedback.
    4. Per-problem -- the most flexible.
3. Specifying hidden vs. revealed feedback
    1. As an additional parameter in the declaration of a problem. It would be a pain to go and change this for every problem type we support.
    2. Via some GUI where instructors could show/hide feedback in the admin interface. This appeals the most to me.
4. Security
    1. For low-stakes assessments -- in-class exercises where you don't want to give feedback until students have already worked the problem; hiding feedback before an assignment is due.
    2. High-stakes assessments -- provide some means to ensure student's can't even see the problems before they begin the exam. I do this my requiring an auth code only TAs know; TAs proctor the exam to help reduce cheating.

So, we need to discuss what makes sense. This PR implements a static, site-wide, high-stakes variant.

bnmnetp: I think there is a lot more to this that we want to consider before moving forward with this PR.

I'd like to see us spend the time actually designing this, either in an issue as I did for the reveal directive or on the wiki.

There are a lot of different use cases and a lot of different levels of security and a number of different deployment scenarios to figure out.

There is also an issue to consider around how much of the timed_exam directive we can leverage.  At a minimum it can benefit from some of this work, and we should not duplicate work that is already done there.

bnmnetp: We may want to consider whether some set of flags applies to feedback around all of the components when in dynamic mode with respect to server side versus client side checking of answers.

bjones1: I agree. Would you like to move this to an issue? Or just use this PR as a place to discuss? (The PR as is doesn't make sense to merge -- whatever it becomes, it will be very different.)

presnick:
One more use case I’d like to plan for:

a problem set where unit tests are used for auto grading but results are
not shown to students.

bjones1:
Good point. Assuming we have the granularity to hide/show feedback from individual questions, the ability to enable/disable by problem type and homework assignment would be handy

models/0.py
bnmnetp: Lets close this and move it to an issue.

PR #1245, opened 2019-06-01
Test improvements

bjones1:
Let web2py define as many tables as possible, and do a better job of recovering from a broken database.

For example, this PR doesn't touch `runestone_test.sql` and tests fail with a plain `python run_tests.py`. However, using `python run_tests.py --rebuildgrades` fixes the database and makes the test succeed.

### DIFF
--- a/tests/run_tests.py
+++ b/tests/run_tests.py
@@ -71,9 +71,9 @@ if __name__ == '__main__':
         print('Skipping DB initialization.')
     else:
         # make sure runestone_test is nice and clean.
-        xqt('dropdb --echo --if-exists --host={} --user={} "{}"'.format(pgnetloc, pguser, dbname),
-            'createdb --echo --host={} --user={} "{}"'.format(pgnetloc, pguser, dbname),
-            'psql  --host={} --user={} "{}" < runestone_test.sql'.format(pgnetloc, pguser, dbname))
+        xqt('dropdb --echo --if-exists --host={} --username={} "{}"'.format(pgnetloc, pguser, dbname),
+            'createdb --echo --host={} --username={} "{}"'.format(pgnetloc, pguser, dbname),
+            'psql  --host={} --username={} "{}" < runestone_test.sql'.format(pgnetloc, pguser, dbname))
         # Let web2py recreate certain tables not in the test database.
         for path in glob.glob('../databases/test_runestone_*.table'):
             os.remove(path)
--- a/tests/run_tests.py
+++ b/tests/run_tests.py
@@ -29,11 +29,13 @@ COVER_DIRS = 'applications/runestone/modules,applications/runestone/controllers,
 
 # Assume we are running with working directory in tests
 
+
 # Functions
 # =========
 # Let web2py recreate certain tables not in the test database by removing their
 # migration metadata.
 def remove_web2py_tables():
+    print('Removing web2py test database migration information (.table files).')
     for path in glob.glob('../databases/test_runestone_*.table'):
         os.remove(path)
 
@@ -136,6 +138,7 @@ if __name__ == '__main__':
     with pushd('../../..'):
         if extra_args:
             print('Passing the additional arguments {} to pytest.'.format(' '.join(extra_args)))
+        xqt('ls -la applications/runestone/databases')
         # Now run tests.
         xqt('{} -m coverage erase'.format(sys.executable),
             '{} -m pytest -v applications/runestone/tests/test_server.py {}'.format(sys.executable, ' '.join(extra_args)),
--- a/tests/run_tests.py
+++ b/tests/run_tests.py
@@ -29,6 +29,17 @@ COVER_DIRS = 'applications/runestone/modules,applications/runestone/controllers,
 
 # Assume we are running with working directory in tests
 
+# Functions
+# =========
+# Let web2py recreate certain tables not in the test database by removing their
+# migration metadata.
+def remove_web2py_tables():
+    for path in glob.glob('../databases/test_runestone_*.table'):
+        os.remove(path)
+
+
+# main
+# ====
 if __name__ == '__main__':
     os.environ['WEB2PY_CONFIG'] = 'test'
     # HINT: make sure that ``0.py`` has something like the following, that reads this environment variable:
@@ -70,18 +81,23 @@ if __name__ == '__main__':
     if parsed_args.skipdbinit:
         print('Skipping DB initialization.')
     else:
-        # make sure runestone_test is nice and clean.
+        # Make sure runestone_test is nice and clean -- this will remove many
+        # tables that web2py will then re-create.
         xqt('dropdb --echo --if-exists --host={} --username={} "{}"'.format(pgnetloc, pguser, dbname),
             'createdb --echo --host={} --username={} "{}"'.format(pgnetloc, pguser, dbname),
             'psql  --host={} --username={} "{}" < runestone_test.sql'.format(pgnetloc, pguser, dbname))
-        # Let web2py recreate certain tables not in the test database.
-        for path in glob.glob('../databases/test_runestone_*.table'):
-            os.remove(path)
+        # Tell web2py that tables have been removed from the database so that
+        # its migration algorithm will re-create them.
+        remove_web2py_tables()
 
         if parsed_args.rebuildgrades:
+            # Paranoia: if the existing ``runestone_test.sql`` database contains
+            # broken tables, remove all tables that web2py can re-create to
+            # return the databased to a fixed state.
+            xqt('psql "{}" < runestone_clean.sql'.format(dbname))
             with pushd('../../..'):
                 print("recalculating grades tables")
-                # TODO: Running this causes the tests to fail.
+                # TODO: This causes test failures when run.
                 #xqt('{} web2py.py -S runestone -M -R applications/runestone/tests/make_clean_db_with_grades.py'.format(sys.executable))
 
             # Copy the test book to the books directory.
@@ -102,8 +118,12 @@ if __name__ == '__main__':
                     '{} -m runestone deploy'.format(sys.executable))
 
             print("dumping the data")
+            # Remove all tables that web2py will re-create before saving the
+            # database. This is the second database clean (the first happens
+            # before recalculating the grades tables), since the recalculation
+            # re-creates the tables.
+            remove_web2py_tables()
             xqt(
-                # Remove all tables that web2py auto-creates before saving the table.
                 'psql "{}" < runestone_clean.sql'.format(dbname),
                 'pg_dump --no-owner "{}" > runestone_test.sql'.format(dbname),
             )
--- a/tests/run_tests.py
+++ b/tests/run_tests.py
@@ -67,15 +67,6 @@ if __name__ == '__main__':
     # Per https://docs.python.org/2/library/argparse.html#partial-parsing, gather any known args. These will be passed to pytest.
     parsed_args, extra_args = parser.parse_known_args()
 
-    if parsed_args.rebuildgrades:
-        with pushd('../../..'):
-            print("recalculating grades tables")
-            xqt('{} web2py.py -S runestone -M -R applications/runestone/tests/make_clean_db_with_grades.py'.format(sys.executable))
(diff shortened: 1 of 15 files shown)


tests/run_tests.py
```diff
@@ -47,37 +78,55 @@
...
+                    '{} -m runestone deploy'.format(sys.executable))
+
+            print("dumping the data")
+            # Remove all tables that web2py will re-create before saving the
+            # database. This is the second database clean (the first happens
+            # before recalculating the grades tables), since the recalculation
+            # re-creates the tables.
+            remove_web2py_tables()
+            xqt(
+                'psql "{}" < runestone_clean.sql'.format(dbname),
```
bnmnetp: I know you said that this was an intermediate step, but I can't figure out what this is all trying to accomplish...

The original goal was to not do migrations with the runestone_test.sql file so that we would know exactly the state of the database.  Yes, this meant we had to do a bit of extra work when a schema change happened, but it allowed us to avoid unpleasant surprises and errors due to migrations.  Remember this was a stopgap measure because we had literally NO unit tests for the server at that time.

I'm having a hard time understanding what state we might be in after all of this:
1. Drop the database which removes everything
2. Reload from the existing runestone_test.sql file
2. remove the table files
3. potentially rebuild then grades
3. drop the tables with the runestone_clean.sql file
4. then do a dump right after you have removed a bunch of tables??

I know this is meant to be an interim solution, but I am trying to figure out what problem this solves or how this makes things better.

In the long run, It seems like a cleaner solution would be to:

1. Let rsmanage init create the database
2. Save the **data** in the current runestone_test.sql to CSV files
3. repopulate the tables from the CSV files using the appropriate API

The big challenge with the unit tests is always setting things up with the appropriate amount of populated course data so that we can test whatever it is we want to test.  In the case of the runestone_test.sql file it was Primarily aimed at testing the autograding mechanism.  

But I struggle with this for every new feature.  OK, I'd like to test feature X but I need an assignment with several questions and several users and this user should be correct for these, etc.  suddenly I need to spend 5 hours writing test code for a feature that only took me a few minutes to add.  We have got to find a way to turn the economics of this around.

bjones1: Commenting on what this PR does and its future - the problem: 
1. `runestone_test_sql` is out of date and needs to be migrated. 
2. web2py refuses to migrate, since its `.table` files are out of sync with the state of `runestone_test.sql`.
3. For old unit tests, we must import data from parts of `runestone_test.sql`

The interim solution: require web2py and runestone to re-create all parts of the database that don't need pre-loaded data for the old unit tests. To do so:
1. Drop and re-create the database itself, so there are no tables.
2. Load in `runestone_test.sql` data.
3. From this, drop all tables that the old unit tests don't rely on and remove the corresponding .`table` files so that web2py knows these should be re-created.
4. Recalculate grades
5. `runestone build` then `runestone deploy`, then copy this from `test/` to `books/`.
6. From step 4, web2py has re-created a bunch of tables that it can also re-create later. Drop all these and the corresponding `.table` files. Save this in a new, corrected `runestone_test.sql` file.
7. Now, the database is in a fully "fixed" state. Run all the tests. On future runs, skip steps 1-6 to save time, instead reloading from `runestone_test.sql` and dropping the corresponding `.table` files.

The long-term solution:
1. Use `rsmanage init` to create the database. Delete all `.table` files.
2. Run `runestone build` and `runestone deploy` to populate it with a test book. Save this to `runestone_test.sql`
3. Run the tests. On future runs, skip steps 1-2, instead reloading the database from `runestone_test.sql`.

bjones1: Sigh. I'm still debugging this -- it runs fine on Windows, but not on Linux.

bnmnetp: I'll give it a try in my docker container.

One thing I don't understand... I just re-ran all the tests on the current master and everything passes... So, what about this PR breaks the current tests to the point it becomes necessary to recalculate all of the grades?

bnmnetp: For me it crashes out almost immediately with an error that the web2py_session_runestone table already exists.  So somewhere in your dropping and recreating of tables this one got overlooked.

bjones1: This PR makes migrations (mostly) not require manual editing of `runestone_test.sql`. So, you won't see a major difference until a commit which changes the schema for a table. As a side benefit, it also halves the size of `runestone_test.sql`.

bjones1: This works fine on Windows. I haven't been table to get Docker running, but I plan on debugging it there when I do.

bnmnetp: I've got a call right now, but I would be happy to help you get docker up and running after 3:30

bnmnetp: I get the same error on my mac as I do on Docker.

```
___________________ ERROR collecting applications/runestone/tests/test_server.py ___________________
gluon/restricted.py:219: in restricted
    exec(ccode, environment)
applications/runestone/models/db.py:29: in <module>
    session.connect(request, response, db, masterapp=None, migrate=table_migrate_prefix + 'web2py_sessions.table')
gluon/globals.py:953: in connect
    migrate=table_migrate,
gluon/packages/dal/pydal/base.py:590: in define_table
    table = self.lazy_define_table(tablename, *fields, **kwargs)
gluon/packages/dal/pydal/base.py:624: in lazy_define_table
    polymodel=polymodel)
gluon/packages/dal/pydal/adapters/base.py:798: in create_table
    return self.migrator.create_table(*args, **kwargs)
gluon/packages/dal/pydal/migrator.py:281: in create_table
    self.adapter.create_sequence_and_triggers(query, table)
gluon/packages/dal/pydal/adapters/base.py:881: in create_sequence_and_triggers
    self.execute(query)
gluon/packages/dal/pydal/adapters/__init__.py:66: in wrap
    return f(*args, **kwargs)
gluon/packages/dal/pydal/adapters/base.py:413: in execute
    rv = self.cursor.execute(command, *args[1:], **kwargs)
E   psycopg2.errors.DuplicateTable: relation "web2py_session_runestone" already exists

gluon/shell.py:174: in env
    run_models_in(environment)
gluon/compileapp.py:600: in run_models_in
    restricted(ccode, environment, layer=model)
gluon/restricted.py:235: in restricted
    raise RestrictedError(layer, scode, output, environment)
E   gluon.restricted.RestrictedError

applications/runestone/tests/test_server.py:171: in <module>
    RUNESTONE_ENV = web2py_controller_env(RUNESTONE_CONTROLLER)
applications/runestone/tests/test_server.py:76: in web2py_controller_env
    env = gluon.shell.env(application, import_models=True)
gluon/shell.py:177: in env
    sys.exit(1)
E   SystemExit: 1
----------------------------------------- Captured stderr ------------------------------------------
  File "/Users/bradleymiller/Runestone/web2py/gluon/restricted.py", line 219, in restricted
    exec(ccode, environment)
  File "applications/runestone/models/db.py", line 29, in <module>
    session.connect(request, response, db, masterapp=None, migrate=table_migrate_prefix + 'web2py_sessions.table')
  File "/Users/bradleymiller/Runestone/web2py/gluon/globals.py", line 953, in connect
    migrate=table_migrate,
  File "/Users/bradleymiller/Runestone/web2py/gluon/packages/dal/pydal/base.py", line 590, in define_table
    table = self.lazy_define_table(tablename, *fields, **kwargs)
  File "/Users/bradleymiller/Runestone/web2py/gluon/packages/dal/pydal/base.py", line 624, in lazy_define_table
    polymodel=polymodel)
  File "/Users/bradleymiller/Runestone/web2py/gluon/packages/dal/pydal/adapters/base.py", line 798, in create_table
    return self.migrator.create_table(*args, **kwargs)
  File "/Users/bradleymiller/Runestone/web2py/gluon/packages/dal/pydal/migrator.py", line 281, in create_table
    self.adapter.create_sequence_and_triggers(query, table)
  File "/Users/bradleymiller/Runestone/web2py/gluon/packages/dal/pydal/adapters/base.py", line 881, in create_sequence_and_triggers
    self.execute(query)
  File "/Users/bradleymiller/Runestone/web2py/gluon/packages/dal/pydal/adapters/__init__.py", line 66, in wrap
    return f(*args, **kwargs)
  File "/Users/bradleymiller/Runestone/web2py/gluon/packages/dal/pydal/adapters/base.py", line 413, in execute
    rv = self.cursor.execute(command, *args[1:], **kwargs)
psycopg2.errors.DuplicateTable: relation "web2py_session_runestone" already exists


!!!!!!!!!!!!!!!!!!!!!!!!!!!!! Interrupted: 1 errors during collection !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
===================================== 1 error in 0.94 seconds ======================================
```

tests/run_tests.py
```diff
@@ -47,37 +78,55 @@
...
+            'psql  --host={} --username={} "{}" < runestone_test.sql'.format(pgnetloc, pguser, dbname))
+        # Tell web2py that tables have been removed from the database so that
+        # its migration algorithm will re-create them.
+        remove_web2py_tables()
+
+        if parsed_args.rebuildgrades:
+            # Paranoia: if the existing ``runestone_test.sql`` database contains
+            # broken tables, remove all tables that web2py can re-create to
+            # return the databased to a fixed state.
+            xqt('psql "{}" < runestone_clean.sql'.format(dbname))
```
bnmnetp: This needs to use the host and username information like the other psql commands.  We cannot assume that the database will always be running on the same host as the tests, and in fact in our Docker configuration the database is in a separate container.

bjones1: Good point -- thanks. I'll update my PR. I'm still seeing some strange Docker things, so I'm rebuilding my VM / Docker setup and hope to debug this soon.

bnmnetp:
Just brainstorming here, but this problem has been gnawing at the back of my brain...

I think we can do a good interim solution if we take table creation out of the hands of the runestone_test.sql file.    That is we 

1.  `rsmanage init` create all the tables. 
2.  Then we simply LOAD that data into the existing tables from runestone_test.sql - This should work fine for the older tests as long as we don't delete a column.
3. Then we build the test book and let that populate
4. At this point we should turn off migrations and run the tests.

It is dangerous and hard to try to outguess and outthink what web2py is going to do with the files in the databases folder.

bjones1:
That seems reasonable. Some questions/thoughts:

1. Before this, all the web2py table files must be deleted (I think).
2. How? Other than psql, which of course creates tables. Unless you propose hand-editing that, which sounds like a lot of work...
3. Yes.
4. Yes.

bnmnetp:
1.  Yes, although `rsmanage initdb  --reset` will take care of dropping the database and eliminating the .table files.
2. You can tell pg_dump to only dump the data and not the schema.

bnmnetp:
@bjones1 I created a data only dump, and hand edited it to remove the rows that are automatically inserted by `rsmanage initdb --reset` - for example row 1 of the auth_groups table, and the standard courses.

bjones1:
Wow, that sounds great! If you haven't already, you might pre-process it be running it through the `runestone_clean.sql` in my PR to drop all tables whose data isn't needed for the tests.

Shall I close this PR, or do you want to take and modify it? I'm excited that you've found a good way forward with this.

bnmnetp:
@bjones1 I have pushed `rtdata.sql` to the master branch of the repo. After doing a `rsmanage initdb --reset` it loads without error.

bnmnetp:
Why don't I take a shot and modifying what you started to use this approach. I don't think it should take long to get things passing..... (famous last words)

bjones1:
Sounds great! All yours!

bnmnetp:
OK, I just merged our combined efforts to master.  But I'll open a PR for some issues with rebuildgrades.  Basically all we were testing before was that rebuild grades did not crash.  
If I run with --rebuildgrades now, I get one test failure:
```
  File "applications/runestone/tests/test_assignments.py", line 252, in test_calculate_totals
    self.assertEqual(res['computed_score'], 5.0)
AssertionError: 10.0 != 5.0
```

Now, we were never meant to run with the rebuilt grades, that was only put in place so that we could respond to some kind of change in the grading were we to discover a bug there.

The problem is that we've been testing so long without actually using rebuilt grades that I have no idea if the problem is in the test code, the code that populates the grades table, or if something changed.  My hunch -- based on a working with people who carefully checked the grades -- is that there is something in rebuildgrades that needs updating.  So, we are no worse off than we were yesterday, but we need to figure out whats causing the problem.

bjones1:
Thanks for working on this!

I agree re: the grade rebuild. I think it's not a big deal, but should be tracked down at some point, Or, perhaps we should just migrate that test away from the old approach then not worry about it.

PR #1251, opened 2019-06-06
Remove code related to cohorts

bnmnetp:
@bjones1 , @presnick -- this removes a lot of code and views related to the cohort feature.  We deprecataed that a couple of years ago and it has essentially no use on either academy or the old interactivepython site.

If we need to resurrect the feature we can do so from this PR later, but for now I'm definitely in favor of cleaning out dead code and simplifying.

Any objections?


bjones1:
Wow! This is fantastic! Go Brad!

A suggestion for more cleanup: there are a huge number of classes defined in `models\db_dashbard.py`, which means these must be imported on **every page load**. Most seem like they could be moved to controllers where they're actually used.

bnmnetp:
At least a year ago, I had a branch that moved db_dashboard.py (WAY out of date now) to the modules folder so that it could be included only as necessary.  I agree that it is crazy to have it in models as that means it gets loaded on every request.

PR #1252, opened 2019-06-06
Misc fixes

bjones1:
More minor cleanup.

### DIFF
--- a/rsmanage/rsmanager.py
+++ b/rsmanage/rsmanager.py
@@ -80,9 +80,9 @@ def initdb(config, list_tables, reset, fake, force):
     if reset:
         if not force:
             click.confirm("Resetting the database will delete the database and the contents of the databases folder.  Are you sure?", default=False, abort=True, prompt_suffix=': ', show_default=True, err=False)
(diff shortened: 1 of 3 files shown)


rsmanage/rsmanager.py
```diff
@@ -80,9 +80,9 @@ def initdb(config, list_tables, reset, fake, force):
     if reset:
         if not force:
             click.confirm("Resetting the database will delete the database and the contents of the databases folder.  Are you sure?", default=False, abort=True, prompt_suffix=': ', show_default=True, err=False)
-        res = subprocess.call("dropdb --if-exists --host={} --user={} {}".format(config.dbhost, config.dbuser, config.dbname),shell=True)
+        res = subprocess.call("dropdb --if-exists --host={} --username={} {}".format(config.dbhost, config.dbuser, config.dbname),shell=True)
```
bnmnetp: I've always used --user and it continues to work for me.  Is this a windows thing?

bjones1: Perhaps -- per the [docs](https://www.postgresql.org/docs/9.6/app-psql.html), that's the official name...

PR #1253, opened 2019-06-06
Dash move

bnmnetp:
Move db_dashboard.py to modules so it is not loaded on every request.

There are challenges with moving something like this to a module in that you need to use the `gluon.current` object to get at request, and db, etc.

The old unit tests are totally unable to cope with this craziness so I've disabled the two unit tests in test_dashboard.py (I've tested them extensively from the browser) and I'm going to write them in the style of test_server.py

@bjones1 my hand testing of these thing uncovered a problem with your utility function `get_course_row` by default it only returns a couple of columns from courses but anything that uses the `_sphinx_static_files.html` template will need much more.  I fixed the calls in dashboard.py to pass `db.courses.ALL` but there may be other problems lurking....

### DIFF
--- a/modules/db_dashboard.py
+++ b/modules/db_dashboard.py
@@ -188,7 +188,7 @@ class UserActivity(object):
         # returns page views for the last 7 days
         recentViewCount = 0
         current = len(self.rows) - 1
-        while current >= 0 and self.rows[current]['timestamp'] >= datetime.datetime.utcnow() - timedelta(days=7):
+        while current >= 0 and self.rows[current]['timestamp'] >= datetime.utcnow() - timedelta(days=7):
             recentViewCount += 1
             current = current - 1
         return recentViewCount
@@ -386,7 +386,7 @@ class DashboardDataAnalyzer(object):
         if not self.user:
             rslogger.debug("ERROR - NO USER username={} course_id={}".format(username, self.course_id))
             current.session.flash = 'Please make sure you are in the correct course'
-            current.redirect('/runestone/default/courses')
+            current.redirect(URL('default', 'courses'))
             # TODO: calling redirect here is kind of a hacky way to handle this.
 
         self.logs = current.db((current.db.useinfo.course_id==self.course.course_name) &
@@ -455,7 +455,7 @@ class DashboardDataAnalyzer(object):
                                                "due_date":assign["duedate"].date().strftime("%m-%d-%Y")}
 
 # This whole object is a workaround because these strings
-# are not generated and stored in the current.db. This needs automating
+# are not generated and stored in the db. This needs automating
 # to support all books.
 class IdConverter(object):
     problem_id_map = {
@@ -479,4 +479,4 @@ class IdConverter(object):
 
     @staticmethod
     def problem_id_to_text(problem_id):
-        return IdConverter.problem_id_map.get(problem_id, problem_id)
+        return IdConverter.problem_id_map.get(problem_id, "DEFUALT NAME USED WHERE??")
--- a/modules/db_dashboard.py
+++ b/modules/db_dashboard.py
@@ -344,7 +344,7 @@ class DashboardDataAnalyzer(object):
     def load_chapter_metrics(self, chapter):
         if not chapter:
             rslogger.error("chapter not set, abort!")
-            session.flash = "Error No Course Data in DB"
+            current.session.flash = "Error No Course Data in DB"
             return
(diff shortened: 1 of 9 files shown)


modules/db_dashboard.py
```diff
@@ -2,11 +2,12 @@
...
 import six
+from gluon import current
 
-rslogger = logging.getLogger(settings.logger)
-rslogger.setLevel(settings.log_level)
+rslogger = logging.getLogger(current.settings.logger)
+rslogger.setLevel(current.settings.log_level)
 
-#db.define_table('dash_problem_answers',
+#current.db.define_table('dash_problem_answers',
```
bjones1: Why not just delete these? Are they used?

modules/db_dashboard.py
```diff
@@ -17,7 +18,7 @@
 #)
 
 
-#db.define_table('dash_problem_user_metrics',
+#current.db.define_table('dash_problem_user_metrics',
```
bjones1: Same here.

modules/db_dashboard.py
```diff
@@ -369,50 +372,51 @@ def load_chapter_metrics(self, chapter):
...
             rslogger.debug("ERROR - NO USER username={} course_id={}".format(username, self.course_id))
-            session.flash = 'Please make sure you are in the correct course'
-            redirect('/runestone/default/courses')
-
-        self.logs = db((db.useinfo.course_id==self.course.course_name) &
-                       (db.useinfo.sid == username) &
-                       (db.useinfo.timestamp >= self.course.term_start_date)).select(db.useinfo.timestamp,db.useinfo.sid, db.useinfo.event,db.useinfo.act,db.useinfo.div_id, orderby=~db.useinfo.timestamp)
-        self.db_chapter_progress = db((db.user_sub_chapter_progress.user_id == self.user.id)).select(db.user_sub_chapter_progress.chapter_id,db.user_sub_chapter_progress.sub_chapter_id,db.user_sub_chapter_progress.status)
+            current.session.flash = 'Please make sure you are in the correct course'
+            current.redirect('/runestone/default/courses')
```
bjones1: Use `URL('default', 'courses')` instead to avoid hard-coding the application.

modules/db_dashboard.py
```diff
@@ -451,7 +455,7 @@ def load_assignment_metrics(self, username, studentView=False):
                                                "due_date":assign["duedate"].date().strftime("%m-%d-%Y")}
 
 # This whole object is a workaround because these strings
-# are not generated and stored in the db. This needs automating
+# are not generated and stored in the current.db. This needs automating
```
bjones1: A bit of overzealous find-and-replace...

tests/test_dashboard.py
```diff
@@ -28,15 +28,21 @@ class TestDashboardEndpoints(unittest.TestCase):
     def setUp(self):
         global request, session, auth
         request = Request(globals()) # Use a clean Request object
+        request.folder = "/srv/web2py/applications/runestone/"
```
bjones1: Looks suspicious -- will only work in a Docker container...

bjones1:
Hmmm, good point re: `get_course_row`. Hopefully, testing will uncover more cases...

tests/test_dashboard.py
bnmnetp: Yes, but I'm going to rewrite it entirely so I don't think it really matters.

modules/db_dashboard.py
bnmnetp: apparently not, since they've been commented out forever.

tests/test_dashboard.py
bjones1: Sounds good.

modules/db_dashboard.py
bnmnetp: I was actually on the verge of deleting them, but then I decided they they went along with the comment that followed.  That much of what we show in dashboard could be calculated and stored in one of the tables represented by the commented out definitions.

I suppose we could just copy them into an issue somewhere, but they aren't hurting anything where they are either.

bjones1: Sounds good.

modules/db_dashboard.py
bnmnetp: This `IdConverter` object requires further investigation... I'm pretty sure its not used, and is insanely incomplete if it is, in fact thethings it is looking for are in the database.

PR #1254, opened 2019-06-09
Removeuser

bnmnetp:
I have been handling user delete requests through email.  This makes it a self service option.  This is in better compliance with the GDPR requirements.


bnmnetp:
@bjones1 -- Something is not quite right.

The test for this PR was passing yesterday before I merged your new PR.  I have verified by hand using a browser that all the apprropriate rows are deleted, and I've even moved the assert statements to the end of the delete function in default.py where they work fine.

I feel like I must need to do something different in how I'm setting up the user or course or something because if I uncomment the asserts at the end of test_deleteacccount it fails.

```
>       assert not db(db.auth_user.username == 'user_to_delete').select().first()
E       assert not <Row {'id': 1, 'auth_membership': <Set ("auth_membership"."user_id" = 1)>, 'auth_event': <Set ("auth_event"."user_id" ...ration_id': '', 'course_id': 14, 'course_name': 'test_course_3', 'accept_tcp': True, 'active': True, 'donated': False}>
E        +  where <Row {'id': 1, 'auth_membership': <Set ("auth_membership"."user_id" = 1)>, 'auth_event': <Set ("auth_event"."user_id" ...ration_id': '', 'course_id': 14, 'course_name': 'test_course_3', 'accept_tcp': True, 'active': True, 'donated': False}> = <bound method Rows.first of <Rows (1)>>()
E        +    where <bound method Rows.first of <Rows (1)>> = <Rows (1)>.first
E        +      where <Rows (1)> = <bound method Set.select of <Set ("auth_user"."username" = 'user_to_delete')>>()
E        +        where <bound method Set.select of <Set ("auth_user"."username" = 'user_to_delete')>> = <Set ("auth_user"."username" = 'user_to_delete')>.select
E        +          where <Set ("auth_user"."username" = 'user_to_delete')> = <DAL uri="postgres://******@db/runestone_test">(<pydal.object...x7f4952826a90> == 'user_to_delete'
E                   -<pydal.objects.Field object at 0x7f4952826a90>
E                   +'user_to_delete')

applications/runestone/tests/test_server.py:514: AssertionError
```

PR #1255, opened 2019-06-10
Fix: Restore initial database state after each test without context anagers

bjones1:
This makes writing tests much easier. It costs ~ 0.3s/test (timings on my Windows PC).


tests/conftest.py
```diff
@@ -158,24 +153,87 @@ def runestone_controller(runestone_env):
...
+    # which should not be truncated is to run ``pg_dump --data-only
+    # $TEST_DBURL > out.sql`` on a clean database, then inspect the output to
+    # see which tables have data. It also excludes all the scheduler tables,
+    # since truncating these tables makes the process take a lot longer.
+    #
+    # The query is:
+    ## SELECT input_table_name || ',' AS truncate_query FROM(SELECT table_schema || '.' || table_name AS input_table_name FROM information_schema.tables WHERE table_schema NOT IN ('pg_catalog', 'information_schema') AND table_name NOT IN ('questions', 'source_code', 'chapters', 'sub_chapters', 'scheduler_run', 'scheduler_task', 'scheduler_task_deps', 'scheduler_worker') AND table_schema NOT LIKE 'pg_toast%') AS information order by input_table_name;
+    db.executesql(
+"""TRUNCATE
+ public.acerror_log,
```
bnmnetp: This seems like a good / fast strategy.  One thing is that initdb does populate some courses and creates an instructor role in the auth_group table.  I'm not seeing those things recreated here?? I'm surprised that the tests would pass without the instructor role in the database.

bjones1: Thanks! The tests create instructors as needed -- I prefer this approach (drop everyone that web2py might modify, including instructors).

bnmnetp: including adding a role to auth_group called instructor ?

bnmnetp:
Based on one successful run, it looks like this model actually does run faster than before.  Thats a good thing as travis was taking 8 minutes sometimes yesterday to run a full suite...

tests/conftest.py
bjones1: Probably not -- the code is just

``` Python
    def make_instructor(self,
        # The ID of the user to make an instructor.
        user_id,
        # The ID of the course in which the user will be an instructor.
        course_id):

        course_instructor_id =  self.db.course_instructor.insert(course=course_id, instructor=user_id)
        self.db.commit()
        db = self.db
        return course_instructor_id
```

Would you suggest a correction? Or should web2py/test code just not touch `auth_group`?

bjones1:
Great! I'm surprised it's faster...

tests/conftest.py
bnmnetp: I guess this is an area where there is some inconsistent use.  In the beginning, we relied on the auth_group table where a user had to be in the instructor group.    So,, a role had to be created in auth_group -- role number 1 was that of 'instructor'

Then we realized that we wanted to have an instructor associated with a specific course, to keep devious students from creating a course just to gain instructor privileges over ALL courses.  (early days!) . Hence the course_instructor table.  when a user creates a course in runestone a role is added to `auth_membership` to indicate they are an instructor and a row is added to `course_instructor` to show they are a member of the instructor group.  `rsmanage addinstructor` does this, but I think your code to create an instructor ignores the `auth_group` and `auth_membership` tables.

bnmnetp: Interestingly, it appears that once we have added a user to the instructor group we never again check that table.  All future checks for instructor verification look at the `course_instructor` table.

bjones1: Sounds like an update to `rs_manage`, then -- don't update `auth_membership`?

bjones1:
Oops, I closed this prematurely -- I thought it was merged to master. If you want me to reopen, let me know.

bnmnetp:
I did merge it to master

PR #1257, opened 2019-06-11
Wip test and report

bjones1:
A mixed of unfinished business:
1. Fixes to the dashboard to correct exceptions seen on my server.
2. Deletion of old files
3. A team feedback system
4. The ability to create an Excel report of an assignment with lots of info about each question. However, it doesn't support all question types.

PR #1260, opened 2019-06-13
Add test_assignment fixture and simple test

bnmnetp:
Create a simple fixture that allows the tester to easily create an assignment, add questions, autograde, calculate totals set deadlines, etc.

Builds on work in #1257

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -90,7 +90,7 @@ def web2py_server(runestone_name, web2py_server_address):
     web2py_server = subprocess.Popen(
         [sys.executable, '-m', 'coverage', 'run', '--append',
          '--source=' + COVER_DIRS, 'web2py.py', '-a', password,
-         '--nogui'],
+         '--nogui', '-e'],
         stdout=subprocess.PIPE, stderr=subprocess.PIPE)
     # Wait for the webserver to come up.
     for tries in range(50):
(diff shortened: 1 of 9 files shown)


tests/conftest.py
```diff
@@ -542,3 +542,100 @@ def test_user(test_client, runestone_db_tools):
...
+        _TestAssignment.assignment_count += 1
+
+
+    def addq_to_assignment(self, **kwargs):
+        if 'points' not in kwargs:
+            kwargs['points'] = 1
+        kwargs['assignment'] = self.assignment_id
+        assert self.test_client.validate(
+            'admin/add__or_update_assignment_question', data=kwargs
+        ) != json.dumps('Error')
```
bjones1: Seems fragile -- is there any way to measure this as an equality (verify that success or something similar was returned)? Here, even a return of `error` would still give a test passed.

tests/conftest.py
```diff
@@ -542,3 +542,100 @@ def test_user(test_client, runestone_db_tools):
...
+        self.due = newdeadline
+        self.save_assignment()
+
+    def save_assignment(self):
+        assert json.loads(
+            self.test_client.validate('admin/save_assignment',
+            data=dict(assignment_id=self.assignment_id,
+                      visible='T' if self.is_visible else 'F',
+                      description=self.description,
+                      due=str(self.due)))) != "ERROR"
```
bjones1: Again, avoid != for tests if possible.

tests/conftest.py
bnmnetp: I copied that straight out of your code.  :smile:

tests/conftest.py
bjones1: Oh no! Do as I say, not as I do... :)

tests/conftest.py
bnmnetp: The success indicator is that a dictionary of values is returned.  I could modify the dictionary to have a status key.  I was trying to keep this code simple and let the actual test take care of details.  For example the keys returned are total, activity_count, and activities_required.  That doesn't seem like something this method ought to worry about.

bnmnetp: OK, I added a simple status key in both cases to indicate success.  No more `!=`

bjones1: Cool!

PR #1261, opened 2019-06-13
Fix: Correctly reference datetime.

bjones1:
I saw this again and tracked it down. It looks like the joy of web2py imports: the last import is apparently `from datetime import datetime`, while `initialize_tables` is using `datetime.date(xxx)`; however, `datetime.date` is a method, not a class, which causes strange-looking errors.

### DIFF
--- a/rsmanage/initialize_tables.py
+++ b/rsmanage/initialize_tables.py
@@ -9,19 +9,19 @@ from os import environ
 
 if db(db.courses.id > 0).isempty():
     click.echo(message="Definining initial Courses", file=None, nl=True, err=False, color='green')
-    db.courses.insert(course_name='boguscourse', term_start_date=datetime.date(2000, 1, 1)) # should be id 1
-    db.courses.insert(course_name='thinkcspy', base_course = 'thinkcspy', term_start_date=datetime.date(2000, 1, 1))
-    db.courses.insert(course_name='pythonds', base_course = 'pythonds', term_start_date=datetime.date(2000, 1, 1))
-    db.courses.insert(course_name='overview', term_start_date=datetime.date(2000, 1, 1))
-    db.courses.insert(course_name='pip2', base_course='pip2', term_start_date=datetime.date(2000, 1, 1))
-    db.courses.insert(course_name='apcsareview', base_course='apcsareview', term_start_date=datetime.date(2000, 1, 1))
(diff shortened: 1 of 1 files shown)


bnmnetp:
I think this means we should look at our imports and make sure we are always importing modules the same way everywhere.

bnmnetp:
Weird, `rsmanage` works perfectly fine for me, all day, every day.  What is going on?

bnmnetp:
I've run that initialize_tables script 100 or more times today in working on tests.

bnmnetp:
Furthermore, the term_start_date is supposed to be just a date, not a timestamp.  And this is part of the `rsmanage` package so it should not be getting loaded anywhere except when the command is run.  It should definitely not be loaded bye the web2py app server process.

bnmnetp:
Are you seeing the problem under windows?  This should not be random. Web2py is supposed to load the files in the models folder in alphabetical order, not randomly.  My usage of datetime.date() Is perfectly fine, and has been that way for well over a year.  So I’m just trying to figure out what the root of this problem is...

bjones1:
I'm running Python 3.7.3 (v3.7.3:ef4ec6ed12, Mar 25 2019, 22:22:05) [MSC v.1916 64 bit (AMD64)] on win32.

However, I can now no longer reproduce this. Earlier, my code wouldn't run without it. Now, my code fails with it. Even more fun, Travis says that this passes. But, of course, web2py doesn't return errors, so https://travis-ci.org/RunestoneInteractive/RunestoneServer/jobs/545484332#L1638 shows that it actually fails.

I'm closing this for now, but I'm also confident that it will show up again.

???

bjones1:
OK, I have a guess on this. The errors happened in my lp branch. In that branch, I import datetime. I'm guessing that import changes the ordering or affects the final result, and hence the breakage.

But this is such a subtle thing -- the imports in one file affecting how another's imports operate. Sigh. That's web2py's design.

bnmnetp:
I think we can limit our exposure if we are more careful and consistent about how we use the datetime module.  

1.  I think we should always just `import datetime` No more `from datetime import datetime`
2. We should make sure that if we use datetime in a file that we import it in that file.  I just noticed that `initialize_tables.py` I am not even importing datetime! That is the fix we should make. I don't think its a good practice to rely on some part of the model or whatever to import stuff for you.  If you use it, import it.

bjones1:
That sounds good.

bnmnetp:
Glad you agree, I just added imports to several files and fixed the import in db_dashboard.

PR #1262, opened 2019-06-20
Instructor page titles

sahi1l:
I've added titles to the instructor's pages so that they can be easily distinguished when opened as tabs.  Hopefully I haven't missed any pages.


sahi1l:
I need to redo this as a clean slate.

PR #1265, opened 2019-06-26
Fix: Correct dashboard exceptions.

bjones1:
This prevents exceptions when looking at the gradebook.

BUT, it should be accompanied with tests that show the failures. Would you have time to work on this, given you've written some autograder tests (and this is hopefully similar)? Hence the draft status -- while this code works, it should have tests.

### DIFF
--- a/controllers/dashboard.py
+++ b/controllers/dashboard.py
@@ -228,10 +228,15 @@ def grades():
     for k in gradebook:
         gradebook[k] = OrderedDict((assign.id,'n/a') for assign in assignments)
 
-    for row in rows:
-        gradebook[row[3]][row[2]] = '{0:.2f}'.format((100 * row[0]/row[1])) if row[1] > 0 else 'n/a'
-        avgs[row[2]]['total'] += (100 * row[0]/row[1]) if row[1] > 0 else 0
-        avgs[row[2]]['count'] += 1 if row[0] >= 0 else 0
+    for score, points, assignments_id, auth_user_id in rows:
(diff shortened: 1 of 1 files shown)


bjones1:
I'm out of time to work on this, unfortunately. However, this does fix exceptions that occur in production code...

bnmnetp:
Just curious, what causes the exceptions?  I've not seen any grade book exceptions in the stream of them over the last week.

bjones1:
As I recall, if a user doesn't work on any problems in an assignment, then it's graded and the report is shown, I see this.

bjones1:
I wonder if this only occurs with LP questions? Perhaps that why I'm the only one seeing it.

controllers/dashboard.py
```diff
@@ -228,10 +228,15 @@ def grades():
     for k in gradebook:
         gradebook[k] = OrderedDict((assign.id,'n/a') for assign in assignments)
 
-    for row in rows:
-        gradebook[row[3]][row[2]] = '{0:.2f}'.format((100 * row[0]/row[1])) if row[1] > 0 else 'n/a'
-        avgs[row[2]]['total'] += (100 * row[0]/row[1]) if row[1] > 0 else 0
-        avgs[row[2]]['count'] += 1 if row[0] >= 0 else 0
+    for score, points, assignments_id, auth_user_id in rows:
```
bnmnetp: Accepting for readability alone 👍 😄

bjones1:
Great, thanks!

PR #1267, opened 2019-07-05
Port of test_ajax.py to test_ajax2.py

psmaker:
All the tests are passing. There is one tests that was not ported:

- `testInstructorStatus()` which tests `verifyInstructorStatus()` - this function is already being tested indirectly in a lot of places - for example `test_GetAggregateResults()`. 

One open issue. This line was not ported. Not sure how to handle this part in the new testing framework.

`# @unittest.skipIf(not is_linux, 'preview_question only runs under Linux.') FIXME
`


bjones1:
@psmaker, thanks for your contributions -- porting to the new test framework is a big help for the project! Nice job.

psmaker:
@bnmnetp I was wondering why the Travis-CI reported an error on one of the tests. I looked at it and it was passing fine in my environment. I noticed you pulled in the change, but was wondering if there is an issue I need to be concerned about.

bnmnetp:
@psmaker It was a timing problem.  In many of our queries we were sorting on timestamp to determine the "last" answer, but on my machine and on travis there were multiple answers with the same timestamp.

I made a change to admin.py so that we sort by id instead and now they pass reliably.  The id increases with each insertion so it is a more reliable indicator.

So... the unit tests did their job!

PR #1269, opened 2019-07-10
Add copy assignment(s) Feature

bnmnetp:
Many instructors have asked for the ability to copy an assignment from a previous course to their current course.  This PR adds that.

- [x] Server side code and tests to do the copying
- [x] Add UI to allow the instructor to request this.  
  - Maybe when course is created 
  - Maybe from the assignments tab


bnmnetp:
@presnick -- check this out.  It allows you to copy all the assignments from a previous course to your new course.

PR #1270, opened 2019-07-10
Cleanup

bjones1:
On the choose course page, the buttons were too narrow for some of my course names. This makes them wider, so students can see what course they're selecting.

There's also a bit of test cleanup -- renames for clarity.

### DIFF
--- a/static/css/dd.css
+++ b/static/css/dd.css
@@ -1,61 +1,61 @@
 #lists {background-color: #aaaaaa;}
 
 #boxdiv {width:400px;height:30px;padding:10px;border:3px solid #aaaaaa; margin-bottom:30px; background-color:#FFCC99}
-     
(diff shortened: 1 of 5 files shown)


bjones1:
I rebased this to resolve conflicts.

static/css/dd.css
```diff
@@ -1,95 +1,97 @@
 #lists {background-color: #aaaaaa;}
 
 #boxdiv {width:400px;height:30px;padding:10px;border:3px solid #aaaaaa; margin-bottom:30px; background-color:#FFCC99}
-
```
bnmnetp: We should publish and adhere to some kind of style guide, so we don't have these semantically meaningless changes introduced into the history.

bjones1: Good point. I missed re-styling this one, unfortunately.

PR #1272, opened 2019-07-11
Wrap stdout and sterr in str() in controllers/ajax.py

sahi1l:
I ran into an error in controllers/ajax.py once because the pipe variables stdout and stderr were concatenated to a string without being converted into a string first. I don't know if this is likely to happen again, but this should prevent it.


sahi1l:
54749e4 should be identical to the current master; I don't know what to do about all of those other commits listed there. Sorry.

sahi1l:
I'm going to close this and do it right. :)

PR #1273, opened 2019-07-11
Grade multiple submissions at once

sahi1l:
In the Grading panel, this lets you select multiple students or multiple questions, and shows the answers for all of them at once. This will let instructors compare multiple answers at once, going back and forth between different students' code without constantly hitting the database (a tedious process).

There are undoubtedly subtle problems with this I haven't thought of, but pleeeeease can we have this for the Fall?

### DIFF
--- a/static/js/admin.js
+++ b/static/js/admin.js
@@ -25,15 +25,15 @@ function gradeIndividualItem() {
     }
 
     else if (VAL3 == 'student') {
-        if (VAL1 == 'assignment' && getSelectedItem('assignment') != null) {
+        var q_column = document.getElementById("gradingcolumn2");
+        sstudents=COL3.selectedOptions;
+        questions=q_column.selectedOptions;
+        if (sstudents.length==1 && (VAL1 == 'assignment' && getSelectedItem('assignment') != null)) {
             calculateTotals()
         } else {
             document.getElementById('assignmentTotalform').style.visibility = 'hidden';
         }
         //we know the question must come from column 2 now
-        var q_column = document.getElementById("gradingcolumn2");
-        sstudents=COL3.selectedOptions;
-        questions=q_column.selectedOptions;
     }
 
     $(rightSideDiv)[0].style.visibility = 'visible';
--- a/static/js/admin.js
+++ b/static/js/admin.js
@@ -204,7 +204,7 @@ function calculateTotals(sid) {
         },
         success: function (retdata) {
             if (retdata.computed_score != null) {
-                // show the form for setting it manually
+                //show the form for setting it manually
                 $('#assignmentTotalform').css('visibility', 'visible');
                 // populate it with data from retdata
                 $('#computed-total-score').val(retdata.computed_score);
@@ -300,7 +300,6 @@ function getRightSideGradingDiv(element, acid, studentId) {
 
         jQuery('#rightTitle', rightDiv).html(data.name + ' <em>' + data.acid + '</em> <span>Points: ' + question_points[data.acid] + '</span>');
 
-        //jQuery('.activecode-target',rightDiv).attr('id',data.acid+"_"+data.username);
 
         if (data.file_includes) {
             // create divids for any files they might need
--- a/static/js/admin.js
+++ b/static/js/admin.js
@@ -475,14 +475,7 @@ function updateColumn2() {
         //FIX: This is where we should get a list of all questions from the chapter
         //chapters[label] should store a list of all question names
         //populateQuestions should be a model for this.
-        $(COL2).empty();
-        for (i = 0; i < chapters[COL1VAL].length; i++) {
-            var option = document.createElement("option");
-            option.text = chapters[COL1VAL][i];
-            option.value = chapters[COL1VAL][i];
-            COL2.add(option);
-        }
-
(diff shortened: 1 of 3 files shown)


bnmnetp:
I'll do a thorough review when I return.  But my initial impressions are that you did some nice refactoring of some really nasty old code. 💯

bnmnetp:
P.S.   -- I don't know if your work fixes this (longstanding bug), but since you are doing major surgery on this code I'll mention it.

We should be able to select a chapter, then an exercise in that chapter, then (one or more) student so that you can easily check work that is not part of an assignment.  Double bonus points for having this working when I'm back on July 20.

sahi1l:
If I knew how to access a list of all the questions in a given chapter, I could do it I think.

bnmnetp:
The questions table in the database has both chapter and sub chapter information.  The name column of the questions table corresponds to the unique I’d of the question.

sahi1l:
I don't know how to access that in Javascript, though.  There is a variable called `assignmentinfo` defined somehow, but I don't know how to create a corresponding `chapterinfo` variable.  (I still have huge gaps in my understanding.)

sahi1l:
I figured out the chapters problem; it's been added to this pull request.

views/admin/grading.html
```diff
@@ -1,7 +1,7 @@
+{{ response.title="Grading" }}
```
bnmnetp: Should this be removed in favor of the PR I just accepted that sets this in the controller?

static/js/admin.js
```diff
@@ -302,7 +307,7 @@ function getRightSideGradingDiv(element, acid, studentId) {
             var file_div_template = '<pre id="file_div_template" style = "display:none;">template text</pre>;'
             var index;
             for (index = 0; index < data.file_includes.length; index += 1) {
-                if (jQuery('#' + data.file_includes[index].acid).length == 0) {
+                if (jQuery('#' + data.file_includes[index].acid,rightDiv).length == 0) {
```
bnmnetp: Is this correct?  What does the ,rightDiv part do?

static/js/admin.js
```diff
@@ -1,65 +1,71 @@
...
-    var select3 = document.getElementById("gradingoption3");
-    var colType = select3.options[select3.selectedIndex].value;
-
-    var col1 = document.getElementById("gradingoption1");
-    var col1val = col1.options[col1.selectedIndex].value;
-
-    var col2 = document.getElementById("gradingoption2");
-    var col2val = col2.options[col2.selectedIndex].value;
+    //This function figures out the parameters to feed to getRightSideGradingDiv, which does most of the work
+    var SEL1 = document.getElementById("gradingoption1");
```
bnmnetp: Is there significance to the all caps variable names?

views/admin/grading.html
sahi1l: Yeah, I don't know how that snuck in there.

static/js/admin.js
sahi1l: ,rightDiv requires that the element be a child of rightDiv; there are now multiple rightDivs, one for each problem to grade, and this is easier than giving all the children of each rightDiv unique names.

static/js/admin.js
sahi1l: Only to make sure I didn't confuse them with the old variables. If this goes against style I can change them to lowercase.

bnmnetp: Makes sense, and although I haven't enforced too much in the way of style, it should be easy enough to make them lowercase now.

I plan to set out some style guidelines and variables should definitely be lowercase.

Thanks!

static/js/admin.js
bnmnetp: gotcha, I've not used that syntax with jQuery before.

bnmnetp:
This should be ready to accept once you get the variable names lowercased!

views/admin/grading.html
sahi1l: Removed from the latest commit.

static/js/admin.js
sahi1l: The latest commit fixes this.

sahi1l:
This should be all set then! Woohoo!

static/js/admin.js
```diff
@@ -409,108 +413,113 @@ function getRightSideGradingDiv(element, acid, studentId) {
...
-
+    else if (val1 == 'chapter' && val2 == 'question') {
+        //FIX: This is where we should get a list of all questions from the chapter
+        //chapters[label] should store a list of all question names
+        //populateQuestions should be a model for this.
+        populateQuestions(col2,chapters[col1val]);
     }
 
-    else if (val == 'student') {
+    else if (VAL1 == 'student') {
```
bnmnetp: You missed a few -- that caused grading to break.  But I took care of them.

bnmnetp:
I merged this from the CL.  Thanks @sahi1l

sahi1l:
This will be so helpful next semester, I hope. Thanks for not making me wait for it. :)

What's the CL?

bnmnetp:
Sure.  I'm happy to add these things that will make life better for you and others.

Command Line

sahi1l:
Oh duh of course, I thought it was some Git jargon I didn't know. :)

PR #1275, opened 2019-07-16
Changes to the Instructor's Dashboard

sahi1l:
Makes the following changes to the dashboard page:

1. Numbers the chapters in the pulldown menu.
2. Adds chapter and section number to the Overall Progression bars
BUG: The section number is incorrect for section names that appear more than once, like Glossary, Exercises, etc. 
3. Separates the "progress rings" by section, sorts numerically by section number, and includes section title.

### DIFF
--- a/controllers/dashboard.py
+++ b/controllers/dashboard.py
@@ -47,7 +47,7 @@ class ChapterGet:
     def ChapterNumber(self,label):
         """Given the label of a chapter, return its number"""
         try:
-            return self.Cmap[label].id
+            return self.Cmap[label].chapter_num
         except ValueError:
             return ""
     def ChapterName(self,label):
--- a/controllers/dashboard.py
+++ b/controllers/dashboard.py
@@ -37,10 +37,25 @@ def index():
     course = db(db.courses.id == auth.user.course_id).select().first()
     assignments = db(db.assignments.course == course.id).select(db.assignments.ALL, orderby=db.assignments.name)
     logger.debug("getting chapters for {}".format(auth.user.course_name))
-    chapters = db(db.chapters.course_id == course.base_course).select()
+    chapters = db(db.chapters.course_id == course.base_course).select(orderby=db.chapters.chapter_num)
     chap_map = {}
+    chapnum_map = {}
+    sub_chapters={}
+    subchap_map={}
+    subchapnum_map={}
+    subchapNum_map={} #Not divided by chapter
     for chapter in chapters:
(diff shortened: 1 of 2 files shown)


sahi1l:
The BUG in #2 has been fixed with commit 9eadce1.  

BUG: I created a class called Get which is a terribly stupid name (that lets me do overly clever things like get.ChapterName etc) and will need to be replaced.

controllers/dashboard.py
```diff
@@ -36,11 +85,10 @@ def index():
...
     course = db(db.courses.id == auth.user.course_id).select().first()
     assignments = db(db.assignments.course == course.id).select(db.assignments.ALL, orderby=db.assignments.name)
+    chapters = db(db.chapters.course_id == course.base_course).select(orderby=db.chapters.chapter_num)
+
     logger.debug("getting chapters for {}".format(auth.user.course_name))
-    chapters = db(db.chapters.course_id == course.base_course).select()
-    chap_map = {}
-    for chapter in chapters:
-        chap_map[chapter.chapter_label] = chapter.chapter_name
+    get = Get(chapters) #yes this is a dumb name
```
bnmnetp: You could add a map to the sub_chapter numbers here with a query like
```
sub_chapters = db(db.sub_chapters.chapter_id == chapter.id).select()
for subc in sub_chapters:
   # build a dictionary of dictionaries where the outer index is the chapter and the inner maps the subchapter label to number?
```

sahi1l: But that's what I do in Lines 40-44, in the Get class definition?

bnmnetp: OK, you said you had a bug in that you couldn't use the correct number for exercises and other subchapters that appear in multiple chapters.  But that does give you the right info to use the correct number.

sahi1l: The latest commit fixed that bug.

bnmnetp:
Scott,

This is what I'm seeing right now in one of my test courses.  The subchapter number is correct, but the chapter number is a bit off.

Screen Shot 2019-07-24 at 9 37 07 AM

sahi1l:
Yeah, that's the ID number, not the chapter number. Didn't make a difference for me because I've only got one book loaded.  Commit a2074eb fixes this I think; check it for me please.

bnmnetp:
will check it right after lunch.  In the middle of debugging self grading.

bnmnetp:
Yep, much better.  And very helpful to have everything separated out by subchapter.

PR #1278, opened 2019-07-21
Allow instructor to rename Assignments on assignments tab.

sahi1l:
1) Allow instructor to rename Assignments on assignments tab, using a modal dialog that is similar to the createAssignment dialog.
2) Prevent two assignments from having the same name in the same course (either when creating or renaming an assignment).
3) Change the label on the "Close" button on the modal dialogs in the assignments tab to "Cancel", for clarity. (I wasn't sure if Close meant "do this and then close" or "close instead of doing this")
4) Change the + and - buttons to "Add" and "Delete" for clarity (and to match the new "Rename" button).

### DIFF
--- a/controllers/admin.py
+++ b/controllers/admin.py
@@ -777,10 +777,14 @@ def createAssignment():
     response.headers['content-type'] = 'application/json'
     due = None
     logger.debug(type(request.vars['name']))
-
     try:
-        logger.debug("Adding new assignment {} for course".format(request.vars['name'], auth.user.course_id))
-        newassignID = db.assignments.insert(course=auth.user.course_id, name=request.vars['name'], duedate=datetime.datetime.utcnow() + datetime.timedelta(days=7))
+        name=request.vars['name']
+        course=auth.user.course_id
+        logger.debug("Adding new assignment {} for course".format(request.vars['name'], course))
+        name_existsQ = len(db((db.assignments.name == name) & (db.assignments.course == course)).select())
+        if name_existsQ>0:
+            return json.dumps("EXISTS")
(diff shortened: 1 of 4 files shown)


bnmnetp:
This is causing a crash in the tests, you can find the traceback in the errors folder.

sahi1l:
So the Python 3.7 test is turning up a bug in the same place, so that didn't fix it.  But could someone check the Python 2.7 test result for me? There seems to be a different bug there that I don't understand.

controllers/admin.py
```diff
@@ -777,10 +777,14 @@ def createAssignment():
...
     due = None
     logger.debug(type(request.vars['name']))
-
     try:
-        logger.debug("Adding new assignment {} for course".format(request.vars['name'], auth.user.course_id))
-        newassignID = db.assignments.insert(course=auth.user.course_id, name=request.vars['name'], duedate=datetime.datetime.utcnow() + datetime.timedelta(days=7))
+        name=request.vars['name']
+        course=auth.user.course_id
+        logger.debug("Adding new assignment {} for course".format(request.vars['name'], course))
+        name_existsQ = len(db((db.assignments.name == name) & (db.assignments.course == course)).select())
```
bnmnetp: FYI, you can do `.count()` instead of `.select()` to get the number of matching rows.

bnmnetp:
Scott works for me after one minor change.

Line 1066:
```
var name = form['rename-name'].value;
```

It was ``form.name.value`` but that made everything get renamed to undefined.

No need for you to change it on your side as I have it changed and it will be part of my commit when I merge.

sahi1l:
Should this be closed then?

PR #1279, opened 2019-07-24
Self grade

bnmnetp:
This addresses the conversation on slack around letting students self grade their assignments.

Notes:

1.  Deadlines are not enforced
2. Instructors should make sure that they pay attention to the grading options for problems that cannot be autograded to make sure they are set to manual.
3. Once the instructor has released the grades for an assignment then self grading is turned off

@presnick This uses a bit of the code that you had in place for coursera, but it was incomplete.

### DIFF
--- a/controllers/admin.py
+++ b/controllers/admin.py
@@ -1474,6 +1474,7 @@ def add__or_update_assignment_question():
     question_type = db.questions[question_id].question_type
     chapter = db.questions[question_id].chapter
     subchapter = db.questions[question_id].subchapter
+    auto_grade = db.questions[question_id].autograde
     tmpSp = _get_question_sorting_priority(assignment_id, question_id)
     if tmpSp != None:
         sp = 1 + tmpSp
@@ -1510,6 +1511,10 @@ def add__or_update_assignment_question():
 
     autograde = request.vars.get('autograde')
     which_to_grade = request.vars.get('which_to_grade')
+    if question_type in ('activecode', 'actex'):
+        if auto_grade != 'unittest':
+            autograde = 'manual'
+            which_to_grade = ""
     try:
         # save the assignment_question
         db.assignment_questions.update_or_insert(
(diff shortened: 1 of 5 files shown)


sahi1l:
I think you mentioned this before, but I assume it will be easier to see when a particular problem doesn't have autograding available, so that we can set it to manual?

bnmnetp:
Yes, the treeview picker now shows what kind of thing each name is and has a checkmark for things that are autogradeable.  But even better, for activecode components, it checks to see if there are unit tests and if not then it will automatically set the grading to manual

controllers/admin.py
```diff
@@ -1510,6 +1511,10 @@ def add__or_update_assignment_question():
 
     autograde = request.vars.get('autograde')
     which_to_grade = request.vars.get('which_to_grade')
+    if question_type in ('activecode', 'actex'):
+        if auto_grade != 'unittest':
```
bjones1: You might add a few comments to the code here -- why is this necessary? I'm sure there's a good reason, but I don't know enough about activecode / actex problems to see why.

controllers/assignments.py
```diff
@@ -342,6 +342,22 @@ def calculate_totals():
     else:
         return json.dumps({'success': False, 'message': "Select an assignment before trying to calculate totals."})
 
+@auth.requires_login()
+def student_autograde():
+    """
+    This is a safe endpoint that students can call from the assignment page
+    to get a preliminary grade on their assignment.
+    """
+    assignment_id = request.vars.assignment_id
```
bjones1: I don't remember my web2py, but what if `assignment_id` isn't defined? Will the code break?

We really need to get Selenium-based unit tests going.

controllers/admin.py
bnmnetp: Its because we never set up the defaults correctly for autograding and the picker.  We just blindly set it to % correct.  This leads to confusing scores of 0 visible to the students rather than the Not Yet Graded message.

controllers/assignments.py
bnmnetp: No, it will not break

PR #1287, opened 2019-08-04
Fix self grade

presnick:
This generalizes the student_autograde function so that, in coursera_mode, it will submit the student's grade.

A test is included for the student_autograde function! @bjones1 may want to turn my writing of a models/.py file into a fixture.

There is also some small refactoring of the autograding and calc_totals code. Some of this was necessary because some calls are passing in student id as a DB row number and some are passing in a string, which requires lookups in different columns in the DB. After many confusions and duplications, I finally decided to just make a single function _get_students that would accept either one.

### DIFF
--- a/controllers/assignments.py
+++ b/controllers/assignments.py
@@ -562,7 +562,7 @@ def get_problem():
         res['file_includes'] = [{'acid': acid, 'contents': get_source(acid)} for acid in file_divs]
     return json.dumps(res)
 
-
+@auth.requires_login()
 def doAssignment():
     if not auth.user:
         session.flash = "Please Login"
@@ -720,7 +720,7 @@ def doAssignment():
                 student_id=auth.user.username,
                 released=assignment['released'])
 
-
+@auth.requires_login()
 def chooseAssignment():
     if not auth.user:
         session.flash = "Please Login"
@@ -755,6 +755,7 @@ def _get_practice_completion(user_id, course_name, spacing):
               (db.user_topic_practice_log.q != -1)).count()
 
 # Called when user clicks "I'm done" button.
+@auth.requires_login()
 def checkanswer():
     if not auth.user:
         session.flash = "Please Login"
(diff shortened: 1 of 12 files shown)


modules/rs_grading.py
```diff
@@ -441,21 +441,29 @@ def _compute_assignment_total(student, assignment, course_name, db=None):
...
+def _get_students(course_id=None, sid = None, student_rownum=None, db=None):
+    print("_get_students", course_id, sid, student_rownum)
+    if student_rownum:
+        # get the student id as well as username
+        student_rows = db((db.auth_user.id == student_rownum)
                           ).select(db.auth_user.username, db.auth_user.id)
-    else:
+    elif sid:
+        # sid is a username; get the matching rownum as well
+        # THIS IS DANGEROUS, if username is duplicated across courses. OK if username is unique server-wide
```
bnmnetp: Usernames are unique server wide, enforced by the database

controllers/assignments.py
```diff
@@ -350,33 +403,60 @@ def student_autograde():
...
-        return json.dumps({'success': False, 'message': "Could not find this assignment -- This should not happen"})
+        if settings.coursera_mode:
+            res2 = _calculate_totals(student_rownum=auth.user.id, assignment_id=assignment_id)
+            if not res2['success']:
+                session.flash = "Failed to compute totals for user id {} for assignment {}".format(auth.user.id, assignment_id)
+                res = {'success':False}
+            else:
+                _try_to_send_lti_grade(auth.user.id, assignment_id)
+    return json.dumps(res)
+    # assignment = db(db.assignments.id == assignment_id).select().first()
```
bnmnetp: Is this experimental?  If we don't want this then lets delete the code.

modules/rs_grading.py
presnick: OK. I will remove the comment about it being dangerous.

controllers/assignments.py
presnick: `_try_to_send_lti_grade` is not experimental. It checks to see whether lti is set up for this server and for this student/assignment pair. If so, it sends the grade.

bnmnetp: OK, then there is no reason to keep the commented out code is there?

presnick: I think the commented out code was removed in the other pull request...

PR #1288, opened 2019-08-04
add decorator to require login in several places

presnick:


controllers/assignments.py
```diff
@@ -487,7 +487,7 @@ def get_source(acid):
         res['file_includes'] = [{'acid': acid, 'contents': get_source(acid)} for acid in file_divs]
     return json.dumps(res)
 
-
+@auth.requires_login()
```
bnmnetp: The decorator is redundant with the if not auth.user check right below.

presnick: Yes, but there are a couple functions that didn't have it, where I added it. And I think we should probably use the decorator everywhere and get rid of the check inside the code.

bjones1: I agree -- this is good DRY coding (use a single-line decorator, instead of multi-line checks).

presnick: I have removed the redundant auth checks now.

PR #1289, opened 2019-08-06
LTI

bjones1:
This PR contains a fix and an improvement:
- Lengthen LTI fields per the spec to avoid internal errors.
- Auto-push the grade when a student browses to a graded, released assignment.


bnmnetp:
I want to make sure that as the two main users of LTI that you and @presnick are on the same page with this change.

bjones1:
Sounds good. I'll wait for Paul's input.

presnick:
I'll try to test this later today.

bjones1:
I've updated this after chatting with @presnick on Slack. Changes:
- Grades are only pushed for a released assignment where a student hasn't visited that assignment yet.
- Use the `_try_to_send_lti_grade` function, refactoring it to the `rs_grading` module.

I didn't change `import oauth` to `import .oauth` (which doesn't work) or something else (which also doesn't work). I think this is fine for two reasons: first, per the [web2py docs](http://web2py.com/books/default/chapter/29/04/the-core#Sharing-the-global-scope-with-modules-using-the-current-object), imports look first in `modules`. Second, there's no third-party OAuth modules I can find on PyPI named `oauth`, except one very old one (updated in 2009) no one is likely to install.

@presnick, would you take another look then approve?

bnmnetp:
@bjones1 -- in our requirements.txt file we include the line `oauth2>=1.9`

Is that a better replacement than the file in modules?

bjones1:
Per [a Slack conversation](https://runestoneteam.slack.com/archives/C0L3XQ0FJ/p1565213422183100), @presnick says that the oauth module was slightly modified by Chuck Severance. Doing a bit of software archaeology (a central plot element in Vernor Vinge's excellent sci-fi novel A Deepness in the Sky), I think:
- [oauth2](https://pypi.org/project/oauth2/), which was last updated in 2015 (yikes), points to a [homepage on Github](https://github.com/joestump/python-oauth2). 
- That page traces it origin from the [original oauth](https://github.com/leah/python-oauth/) repo, which looks similar to the code we're using. This code was last updated in 2012 (!!!). I just finished porting the `oauth.py` code we have to Python 3. 
- While `lti.py` uses `oauth.py`, `rs_grading` imports and uses `OutcomeRequest` from `outcome_request.py`, which uses `oauth2`. So, it seems oauth2 can be used instead of oauth to talk to LTI.

My conclusion: yuck. What a mess. It would be nice to replace `oauth` with `oauth2`, so we don't have to maintain `oauth`. It would also be nice to replace `oauth2` with something actually maintained, such as `oauthlib`. Sigh. But for now, even though there are only about 10 lines of oauth in `lti.py`, I don't know how to translate these easily.

bjones1:
Aargh. I spent too much time and ported the LTI code to oauth2.

bjones1:
Again, the failure is due to Coveralls, not broken tests.

bnmnetp:
Is this good to merge??  You and Paul are both happy with it?

I'm happy to be able to not have to worry about oauth in the modules folder.

bjones1:
I'm happy with it. @presnick, is this OK with you?

presnick:
Looks good, but I can’t test now. OK to merge. I will test next week and
propose any additional changes as needed.

PR #1291, opened 2019-08-06
Whitespace clean

bjones1:
@bnmnetp, do you want this sort of thing fixed? Or is this just annoying?

Doing so shows at least one mis-balanced file (`index.html`) that I don't know how to fix.

IMHO, we should standardize and try to enforce a whitespace policy, especially for web2py templates. Otherwise, they are very hard to read and fix. My proposal:

1. All indents are 4 spaces. Tabs are not allowed.
2. For HTML tags, two styles are supported: `<tag>`Contents`</tag>` (one line) or `<tag>`, then multiple indented lines, then an un-indented `</tag>`.
3. All web2py commands that produce nesting require an indent: `{{ if x: }}`, followed by indented text, then an un-indented `{{ pass }}`.
4. All web2py commands will follow the format: `{{ commands }}`, with one space between the curly braces and the command. For output, use `{{= expression }}`, with one space after the equal sign and one before the closing curly brace.

### DIFF
--- a/views/dashboard/exercisemetrics.html
+++ b/views/dashboard/exercisemetrics.html
@@ -22,26 +22,6 @@
 			</div>
 			<div class="dash-section-content">
 				<div id="chart-exercise-attempts"></div>
-				<!--
-					<div><b>Students who...</b>
-						<div>
-							<div style="height: 10px; width: 10px; border: 1px solid #333; background-color: #00CC66; float: left; margin-top: 3px; margin-right: 3px"></div>
-							got it correct on the first attempt: 13
-						</div>
-						<div>
-							<div style="height: 10px; width: 10px; border: 1px solid #333; background-color: #00CC66; float: left; margin-top: 3px; margin-right: 3px"></div>
-							got it correct after multiple attempts: 15
-						</div>
-						<div>
-							<div style="height: 10px; width: 10px; border: 1px solid #333; background-color: #00CC66; float: left; margin-top: 3px; margin-right: 3px"></div>
-							never got it correct: 3
-						</div>
-						<div>
-							<div style="height: 10px; width: 10px; border: 1px solid #333; background-color: #00CC66; float: left; margin-top: 3px; margin-right: 3px"></div>
-							never attempted it: 2
-						</div>
-					</div>
-				-->
 			</div>
 			<script type="text/javascript">
 				DashboardCharts.exerciseAttempts({{response.write(attempt_histogram,escape=False)}});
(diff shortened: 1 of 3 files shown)


bnmnetp:
I agree that it needs to be fixed, but it is super annoying in that it destroys the history in terms of understanding when the last substantial change was made to the file.

For example, today I was trying to debug something in in books.py and my editor shows me the last time each line was changed.  I was looking at code I wrote, but it said that you had changed it. So I was trying to figure out what you changed and why,   But you really didn't change anything other than an un-indenting of a big block.  I know that is going to happen from time to time when we refactor stuff, but its nice to minimize it.

Personally I would be in favor of working up a style guide for html, python and Javascript code.

On the skulpt project we enforce that by running jshint against the source in our automated build process.  failing a pass by jshint is just like failing a unit test.

bnmnetp:
Regarding your proposal:

1.  Agree
2. Agree
3. I don't feel strongly about this.  It makes the final result inconsistent, but makes it easier for humans to edit.  So I'm somewhat in favor
4. The overwhelming majority of our code uses the convention of {{=var}} (593 lines) very few places follow the {{= expression }} convention you propose.  So unless there is good reason to change I think we should stick with the majority to avoid unnecessary changes to the code and history.

bjones1:
It would be good to have a style guide. But, without a tool to enforce this, it's not that helpful. I know Python code can be linted easily. I'd be in favor of turning this plus jshint for runestone. Unfortunately, I don't know of any way to check web2py templates, since simple HTML checks aren't enough.

I'm happy to change item 4 to agree with most of the code.

bnmnetp:
I think we should still document the style we want for html even if there isn't a good html linter.

It would be a big effort but then we ought to move toward getting a standard set of rules and enforcing both javascript and python lint checking as part of our test suite.

bjones1:
Should styles guide requirements go on the wiki?

In another project I help with, I can copy Travis code to check Python style using flake8 -- see https://github.com/pyinstaller/pyinstaller/blob/develop/.travis.yml#L77. I assume you have similar code to check Javascript?

views/dashboard/exercisemetrics.html
```diff
@@ -22,27 +22,29 @@ <h3>{{=exercise_label}}</h3>
 			</div>
 			<div class="dash-section-content">
 				<div id="chart-exercise-attempts"></div>
-				<!--<div><b>Students who...</b>
```
bnmnetp: Everything from here to line 42 is commented out and has been for 3 years.  I propose we eliminate stuff that has been commented out.  The html is complicated enough with out cluttering it up with stuff that nobody even remembers why it's there.

bjones1: Yes! With pleasure!

(I hesitate to do this on my own, since this isn't my code...)

PR #1295, opened 2019-08-13
Fix: Ensure subprocess output is text.

bjones1:
This came up while debugging changes on the server.


bjones1:
Note that the failing check is due to Coveralls being unable to report stats, not because of test failures.

PR #1298, opened 2019-08-20
Run Validation locally

bnmnetp:
replace calling the w3 service with running the validation locally.

- a bit ugly since everything that uses the w3 validator (vnu.jar) requires java!
- Errors are captured on stdout

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -47,6 +47,7 @@ from html5validator.validator import Validator
 # -------------
 from .utils import COVER_DIRS, DictToObject
 
+# Set this to False if you want to turn off all web page validation.
 W3_VALIDATE = True
 
 # Utilities
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -310,10 +310,12 @@ def _html_prep(text_str):
 
 
 # Create a client for accessing the Runestone server.
+@pytest.mark.usefixtures("tmp_path")
 class _TestClient(WebClient):
-    def __init__(self, web2py_server, web2py_server_address, runestone_name):
+    def __init__(self, web2py_server, web2py_server_address, runestone_name, tmp_path):
         self.web2py_server = web2py_server
         self.web2py_server_address = web2py_server_address
+        self.tmp_path = tmp_path
         super(_TestClient, self).__init__('{}/{}/'.format(self.web2py_server_address, runestone_name),
                                           postbacks=True)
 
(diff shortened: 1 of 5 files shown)


tests/conftest.py
```diff
@@ -37,15 +38,16 @@
...
 import six
 from six.moves.urllib.error import HTTPError, URLError
 from six.moves.urllib.request import urlopen
+from html5validator.validator import Validator
 
 # Local imports
 # -------------
 from .utils import COVER_DIRS, DictToObject
 
+W3_VALIDATE = True
```
bjones1: I'd recommend adding a brief comment to explain what this does. Another handy option would be to add a command-line flag to disable it.

tests/conftest.py
```diff
@@ -347,23 +349,16 @@ def validate(self,
...
-                        if vld.warnings:
-                            print('Warnings for {}: {}'.format(url, len(vld.warnings)))
-                            pprint(vld.warnings)
-                        done = True
-                    except HTTPError as e:
-                        count -= 1
+            if expected_errors is not None and W3_VALIDATE:
+                # Redo this section using html5validate command line
+                vld = Validator(errors_only=True)
+                tmpname = 'tmphtml.html'
```
bjones1: I would recommend using [tmp_path](http://doc.pytest.org/en/latest/tmpdir.html) to create a directory in which to store `tmphtml.html`. I'd rather not have temp files in the local path, even if the code removes them. (Note that pytest will eventually delete these directories, per https://github.com/pytest-dev/pytest/issues/1120).

tests/conftest.py
```diff
@@ -347,23 +349,16 @@ def validate(self,
...
+                with open(tmpname, 'w') as f:
+                    f.write(self.text)
+                errors = vld.validate([tmpname])
+                os.remove(tmpname)
```
bjones1: Remove. Pytest will take care of this.

tests/conftest.py
bnmnetp: I didn't see it as any different than what the other validator did with leaving files in the web2py root folder.  If they are temporary and pytest cleans them up then why add the extra code to put them somewhere else??

tests/conftest.py
bnmnetp: Yep.

tests/conftest.py
bnmnetp: Although now that I look at the docs, it seems that tmp_path / 'tmpfile.html'   Is probably a net zero change.

bnmnetp: but no idea how you incorporate a fixture into code inside a class in conftest.py

bjones1: I agree that my earlier code putting files in the web2py root folder is sloppy and should also be changed.

To incorporate this, my sketch of the code is:

```
# Create a client for accessing the Runestone server.
class _TestClient(WebClient):
    def __init__(self, web2py_server, web2py_server_address, runestone_name, tmp_path <-- added this):
        self.web2py_server = web2py_server
        self.web2py_server_address = web2py_server_address
        self.tmp_path = tmp_path # <-- added this
```
Then modify the fixture just after it to take `tmp_path` and pass it to the constructor.

PR #1300, opened 2019-08-21
Fix: Avoid formatted strings to allow use of Python 3.5.

bjones1:


controllers/books.py
```diff
@@ -21,6 +21,8 @@
 import json
 import logging
 import datetime
+import importlib
```
bnmnetp: If you want to add a comment to this PR explaining our use of importlib that would be great.

bjones1:
OK, done. I also added an `__init__.py` to avoid some of the namespace weirdness.

PR #1318, opened 2019-08-29
Blacken all python code for consistency in formatting

bnmnetp:
This PR reformats EVERYTHING using the black tool.  

It also adds a test to the test suite that will fail if code does not follow the correct format.

Its easy enough for developers to conform if they have black installed.  For example editors like VSCode will automatically run black for you on save. Or when the test fails you simply run black on that file.

This cleans up a massive amount of lint, and should let us focus on cleaning up real errors instead of formatting errors.


bnmnetp:
It seems the black tests pass when I run `pytest -k black` but are failing when they run along with all the rest of the tests.  The issue is a current working directory problem not an actual formatting problem.  But I'll get it cleaned up.

bnmnetp:
OK, tests pass and I've been using this reformatted code on my dev server without any ill effects.  We should get this merged as soon as we can so we can get back to normal development.  I just want and LGTM from both of you to make sure we are all aware of whats going on.

presnick:
It's OK with me if you merge this. I took a quick look and it seemed like everything was fine.

bjones1:
I switch to the blackened version and it looks good! Thanks, Brad.

bnmnetp:
Great!  Done.

There are still some errors that flake8 finds, that black does not deal with.  These are errors like bad variable names. bad escape sequences, the use of bare except statements, multiple imports on one line... These are all things we can easily pick off and clean up.

I've uploaded a .flake8 file that defines a bunch of extra builtins to help with web2pys global variables, sets the line length to match black and a couple of other minor things.

bjones1:
Brad, great work! Having well-formatted code will be so helpful. I also appreciate a flake8 setup -- it will be good to squash those violations as well.

Sigh -- web2py -- there will still be lots of undefined names coming from the automatic import of everything in `models/`. Will we ever move to another web framework (perhaps https://github.com/web2py/py4web)?

PR #1322, opened 2019-09-01
Pr5 lti fix

presnick:
Redirecting and redirecting back doesn't work because requests.vars not set as `lti.py` needs them on the second redirect. 

Eliminating both redirects; students no longer asked to donate as part of account creation workflow in lti.


bjones1:
This will break my payment flow, unfortunately; it also skips the donate step, which I hesitate to do to at leave give Brad a chance to see some income. I'd prefer something that visits the donate/pay pages, but correctly preserves `request.vars` in the process. I'd guess that storing `request.var` in `session` then restoring it at the top of lti would work.

presnick:
@bjones1 I think we need to resolve this, and it should be resolved in favor of not breaking the flow for people who don't want to force visitors to be asked to make donations.

If you want to add a special environment variable flag for your server installation that redirects everyone to make this request, that's OK, but I don't think it should be part of the general flow for all LTI interactions.

bjones1:
I agree -- here's my proposal.

1. Fix the redirect, so that new LTI students visit the donation page.
2. Add an option so that the donation page can be skipped entirely.

Does that sound reasonable?

On a side note, we really need some tests for LTI. Which bring up the larger issue of how well we want to support this. I think good support would increase Runestone's adoption, since it makes it easy to plug in to a LMS and just use.

presnick:
Following the logic we’ve used elsewhere, whatever is deployed on runestone
academy should be the default, and other behaviors should be controlled by
environment variables.

Is LTI available at all on runestone academy? I think that further
development would be needed to think that through. And I think it’s likely
that he would end up making it a premium service that a school has to pay
for rather than individual students. I’m really skeptical about asking
students to donate before they can do their first sign-on.

If this is what Brad has decided on for runestone academy I’m OK with it,
and will use one of my existing environment variables to bypass it. If not,
I think we should go back to my version as the default.

bnmnetp:
LTI is definitely not the default or even remotely in use on Academy.  As far as I can tell there is a lot of work to do to get it to the point where an average instructor could configure something to make it work with however they have their LMS installed.  As Paul said, I would consider this to be a premium feature when we do get to the point where configuration is possible.

I think it should be configurable in the mode like Paul and Bryan run in where you run your own servers for your institution.

presnick:
OK. So there is no natural default then.

My version is working for me. Is your version working for you, Bryan? (My
previous notes on this PR suggest there is a bug.) If your is working for
you, I will make mine controllable by an environment parameter. If not, I
think we should make mine the default and do the debugging as part of a
larger project to make LTI work for everyone.

bjones1:
Yes, my version works for me, with one bug -- if a user tries to navigate to an assignment on the first time they use the system (at registration), they're instead directed to the book instead of the assignment page.

How about this, as a simple solution and as a way to avoid introducing another config variable: only go to the payment/donate page if the price of the book is truthy (meaning it's a paid book)? This would automatically skip the donation page, and only ask for payment if it's absolutely necessary. I think it's just an extra condition on line 180 of the current code. I'd be happy to put together a PR if this seem reasonable.

presnick:
I think it should be a property of the server installation and not the book, or at least not *only* the book.  One possibility is to use the existing settings.academy_mode flag. When it's True, make it work your way; when it's false my way. I already run my servers with academy_mode set to False.

It's not quite as simple as one line of code: if we are not in academy_mode, we need to create the user_courses row even if you're not an instructor, as in my PR.

I suppose one possibility is to put the flag check at line 180, as you suggest. If academy_mode and no user_courses row, then redirect. If not academy_mode, then insert_or_update the user_courses row. That would duplicate the insert_or_update call in the case where the user is an instructor, but that's a rare occasion so the duplicated call isn't a big deal.

Would you like to make a PR along these lines?

bjones1:
I'll be happy to. I'm current buried in beginning of semester tasks, but will return to this when things settle down.

bjones1:
Given that #1411 was merged, this can be closed, correct?

PR #1336, opened 2019-09-13
Fix reading order

bnmnetp:
This fixes a long standing issue with the assignment builder adding chapters in a seemingly random order.

### DIFF
--- a/controllers/admin.py
+++ b/controllers/admin.py
@@ -2113,6 +2113,10 @@ def add__or_update_assignment_question():
                     question_type
                 ],
                 status="success",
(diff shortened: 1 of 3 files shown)


bjones1:
Wow! Nice work. That was a lot more complicated than I expected.

Did you want any help writing Selenium tests?

bnmnetp:
Thanks, it was a good challenge and I definitely feel ready to talk about async/await in class in a few weeks!

I think we need some kind of plan for adding selenium tests.  Certainly tests for assignment creation and assignment grading would be valuable.

PR #1343, opened 2019-09-17
Fix: Correct incorrect subprocess.run parameters introduced in ...

bjones1:
bacb396faa524a600ecdfe5882eec9c3e60a2b1a.


bjones1:
This (finally) works now.

PR #1344, opened 2019-09-20
Misc fixes

bjones1:
This contains updates for my literate programming directive:

1. Use celery instead of web2py's unreliable scheduler.
2. Use Python 3 for subprocess calls.

I've also updated the Docker files so it runs correctly there.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,6 +4,8 @@ python:
 - '3.7'
 services:
 - postgresql
+# For Celery (web2py scheduler replacement).
+- redis-server
 addons:
   postgresql: '9.4'
   apt:
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -8,6 +8,9 @@ services:
       POSTGRES_USER: 'runestone'
       POSTGRES_DB: 'runestone'
 
+  redis:
+    image: redis
+
   runestone:
     image: runestone/server
     restart: always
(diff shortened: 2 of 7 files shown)


.travis.yml
```diff
@@ -4,6 +4,8 @@ python:
 - '3.7'
 services:
 - postgresql
+# For Celery (web2py scheduler replacement).
+- redis-server
```
bnmnetp: interesting.... Once this is in place we can move to using redis to manage session data instead of Postgres.

bjones1: Hmmm - interesting. But I don't see any way to do that -- session data gets stored to either a file or a DB table, and DAL doesn't support redis AFAIK. Or am I missing something? (I hope so -- that sounds really neat...)

bnmnetp: http://web2py.com/books/default/chapter/29/13?search=cache#Sessions-in-Redis

bnmnetp: web2py also supports using redis in places where you would also use memcache.

bjones1: Wow, cool! Nice find!

bjones1:
Pinging...

bnmnetp:
Hey, its on my list of things to test, but I’m not going to get to it until next week.

bjones1:
OK, no problem. I was just afraid it had fallen off your list.

PR #1352, opened 2019-09-30
Misc cleanup

bjones1:
A set of tweaks and cleanup to the tests.


bnmnetp:
If we are removing the start script then you need to update the README with new instructions for how to start a server.

bjones1:
I'm mainly removing this since it's outdated -- it tries to start the scheduler, which has been used only by me (AFAIK) ever since we moved to dynamic books. Would you recommend just editing the script instead? Does anyone actually use this?

bjones1:
For simplicity, I just removed the portion of that script that started the scheduler, but left the rest intact.

bnmnetp:
Thats good.  Thanks.

PR #1359, opened 2019-10-07
plug holes in session expirations

bnmnetp:
This addresses issues where the session cookie times out but the page is not reloaded.  So the eBookConfig.isLoggedIn value is true, but the session cookie has expired.

This also addresses the issue where you have multiple tabs open and change courses in one tab but come back to a previous tab.  Now your current course for the session does not match the course of the book you are displaying.

This also restores the commented out functionality of the idle timer reloading the page, which does seem to work nicely and forces the user to login again when the session expires.

This needs the support of the companion PR on RunestoneComponents

@bjones1 - The most confusing case is when a student logs in to course "babyfopp" a child course of fopp.  Then makes a bookmark -- But the bookmarked url is for books/published/fopp instead of books/published/babyfopp  This allows the student to access fopp.  -- So I'm trying to make it highly apparent that they are in the wrong place by graying out the progress bar.  The mark complete button is already hidden in this case.  But I wonder if there isn't something more clever we could do with url rewriting to make an initial bookmark with the correct course in the path.

### DIFF
--- a/controllers/ajax.py
+++ b/controllers/ajax.py
@@ -45,6 +45,13 @@ def compareAndUpdateCookieData(sid: str):
 def hsblog():
     setCookie = False
     if auth.user:
+        if request.vars.course != auth.user.course_name:
+            return json.dumps(
+                dict(
+                    log=False,
+                    message="You appear to have changed courses in another tab.  Please switch to this course",
+                )
+            )
         sid = auth.user.username
         compareAndUpdateCookieData(sid)
         setCookie = (
@@ -53,6 +60,10 @@ def hsblog():
         # log entries that come from auth timing out even but the user hasn't reloaded
         # the page.
     else:
+        if request.vars.clientLoginStatus == "true":
+            logger.error("Session Expired")
+            return json.dumps(dict(log=False, message="Session Expired"))
+
         if "ipuser" in request.cookies:
             sid = request.cookies["ipuser"].value
             setCookie = True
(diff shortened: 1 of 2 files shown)


bnmnetp:
@bjones1 Here's an idea that will work for the vast majority of students I think.

We already have a runestone cookie that has a longer life of 90 days, so it doesn't expire like a session cookie does.

We could store their 'last class' in that cookie.  If they attempt to access the base class for their last class we could send them to the login page.

@presnick I assume that this PR doesn't really affect you as everyone always has to login for all of the courses you serve.

bjones1:
As I understand it, the problem is that the course babyfopp still has the URL of the base course, `books/published/fopp`. So, if their session cookie expires then they follow this URL, they get taken to the wrong place. One solution would be to provide a more descriptive URL. Of course, we can't use `books/published/babyfopp` since all the images/static files wouldn't be found. So, the solution is either a URL with a query parameter like `books/published/fopp?course=babyfopp` or (as this PR provides) a longer-lasting cookie that will store the current course and provide that, even if the session cookie expires?

bnmnetp:
Right,

In my discussions with a few teachers most of them have the students bookmark the page for the course.  

If the session cookie is still alive then its not a problem.  The router handles that, looks up their current course and takes them to it.

If the session cookie has expired, this PR looks at a cookie that has a. 90 day expiration period  and checks to see if the base course for the course in the cookie is the same as the base course in the URL.  If that is true then they are redirected to login.  I think this close a lot of problems where students are accidentally accessing the base course and "losing work"

bjones1:
Sounds good to me! (The perhaps obvious exception -- exceeding the 90-day limit. But, then, how often will this happen?)

bnmnetp:
The cookie is actually refreshed with a new 90 day expiration whenever they access the course.  So it would be a case where it had been a whole semester since they last accessed the book.

Issue #1259, opened 2019-06-12
Unit Test endpoints -- Help write one test


bnmnetp:
For anyone wanting to jump in and help us get our test coverage up, here is a list of endpoints that could really use some unit tests.  If you have never written a unit test before don't worry, this is a really low stakes way to help out the project, since you are not doing anything that affects the production server!  After all you are not going to cause the server to fail (except if you write a really good test that identifies when the server would fail anyway!)

Here is a list of functions that need unit tests! You can claim just one and submit a PR.
To help you get started check out the documentation here:  https://github.com/RunestoneInteractive/RunestoneServer/tree/master/tests

## admin.py

- [x] assignments
- [ ] practice
- [ ] add_practice_items
- [x] removeStudents
- [x] removeInstructor
- [x] deletecourse
- [x] getQuestions
- [x] questionBank
- [x] getQuestionInfo
- [ ] edit_question
- [x] question_text
- [x] searchstudents
- [x] gettemplate
- [x] createquestion
- [x] htmlsrc
- [x] getStudentCode
- [x] coursename
- [x] releasegrades
- [ ] get_assignment_release_states
- [ ] get_assignment
- [x] save_assignment
- [ ] delete_assignment_question
- [ ] reorder_assignment_questions 
- [x] ~checkQtype~ Not used -- removed

## ajax.py

- [x] ~savegrade~ No longer used.  Remove `listassignments`, `listassessments`, `assessdetai`l and `gradeassignment`
- [ ] checkTimedReset
- [ ] Many many tests do not check for the exception blocks.

## assignments.py

- [ ] record_assignment_score
- [x] autograde
- [ ] record_grade
- [ ] get_problem
- [x] doAssignment (user not logged in)
- [x] chooseAssignment (user not logged in)
- [ ] checkanswer
- [x] practice (user not logged in)
- [x] practice (after course configured for practice)
- [ ] like_dislike
- [ ] practice_feedback

## books.py

- [x] published (for open book when not logged in)

## dashboard.py

- [ ] grades (with practice set up)
- [ ] questiongrades
- [ ] exercisemetrics

## default.py

- [ ] user registering for a course with a prepopulated url
- [ ] coursechooser
- [ ] removecourse
- [ ] sendreport

## designer.py

- [x] build
- [x] ~delete_course~ this was an older duplicate of the function in admin

## proxy.py

- [ ] jobeRun
- [ ] jobePushFile
- [ ] jobeCheckFile

[...]

PR #1364, opened 2019-10-14
Add new tests in test_admin.py

santos22:
See https://github.com/RunestoneInteractive/RunestoneServer/issues/1259

Running tests:
```
$ scripts/dtest -k test_get_assignment_release_states

$ scripts/dtest -k test_delete_assignment_question

$ scripts/dtest -k test_reorder_assignment_questions
```


bnmnetp:
Thanks! 

All tests pass after running `black tests/test_admin.py`. Black is the tool we use to enforce a consistent style on all of our python code.

Many editors allow you to run `black` on save. or you can install it with pip and run it from the command line before you submit a PR.

PR #1384, opened 2019-11-21
fix(1327): No longer duplicates enrollment

blaise2s:
When a student is removed from a course they are no longer moved into the base course if they are already enrolled in it. Thus, this prevents duplicate courses showing up in the Courses page after removing students from classes. The existing functionality of moving a student into the base course if they are **not** enrolled in it still holds true. So, in the case that a student is removed from a course, but they are not enrolled in the base course, they will be moved into that course.


bnmnetp:
This looks good Blaise -- If you want to help with some tests, writing a unit test for this one would be a good place to start.

PR #1387, opened 2019-11-26
Docs: Add manual debug instructions.

bjones1:


tests/README.rst
```diff
@@ -252,3 +252,18 @@ From the scripts folder, run the command:
 
 
 Then in your browser go to `http://127.0.0.1:8089` You an set up how many users you want and how fast they will come online.  The webpage will update every couple of seconds to show you statistics on load times for various kinds of pages.
+
+
+Manual debug
+============
+In order to debug the web2py server and set breakpoints, web2py provides an integrated debugger. However, I'd found it to be unreliable. For a more traditional debugging approach, it's possible to invoke functions defined on the server by creating a web2py controller environment. To do so, invoke Python from the web2py directory then execute the following code:
+
+.. code:: Python
```
bnmnetp: How is this different than running:

`python3.7 web2py.py  -S runestone -M` from the command line?

bjones1: Wow. I honestly didn't know about that. I'll update the readme to use this. Thanks!

bnmnetp: Sorry for not telling you sooner, I just assumed you knew.  I use it a lot for debugging DAL queries.

bjones1: That's often what I'm doing as well. I've updated with your command line instead.

bnmnetp: So if you want to test code in a particular controller do you do something like:

`exec(open('applications/runestone/controllers/dashboard.py').read())` to load the functions?

It may be worth adding a line like that

bjones1: Yes? I haven't tried that, though. The [docs](http://web2py.com/books/default/chapter/29/04/the-core#Command-line-options) also say for the `-S APPNAME` option that "APPNAME like a/c/f (c,f optional)", suggesting this is another way to load a controller?

bjones1: OK, I've experimented with this and updated the instructions. It should be ready to go.

PR #1388, opened 2019-11-26
Fix: Record id of graded row.

bjones1:
This is a first step towards the assignment table, which needs to be able to find which row was graded then display that to the user.

### DIFF
--- a/models/grouped_assignments.py
+++ b/models/grouped_assignments.py
@@ -60,7 +60,8 @@ db.define_table(
     Field("sid", type="string", notnull=True),
     Field("course_name", type="string", notnull=True),
     Field("div_id", type="string", notnull=True),
-    Field("useinfo_id", db.useinfo),  # the particular useinfo run that was graded
+    # The ID of a record in the answer table (mchoice_answers, fitb_answers, etc.) that was graded.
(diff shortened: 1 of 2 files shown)


bnmnetp:
Interesting that you should submit this today.   Yesterday I spent a good chunk of time figuring out why requests by students to delete themselves (required functionality by GDPRs and some US States) were timing out and failing.

It turns out that these foreign key constraints cause delete times to be HUGE on large tables.  For one students with just 500 records in the useinfo table it took over 3 minutes to delete those records.  FYI:  On Runestone.academy the use info table will cross the 100 million rows mark by the end of the week.

So -- I was heading in the opposite direction of you and looking to eliminate these references if they are not absolutely necessary.

Can you describe the use case for adding this reference?

bjones1:
Note that the new field is an integer, so there's no foreign key constraint. I don't think will impact performance.

Use case: I want to show instructors the student's answer that was actually graded, which can be complex to find given the first answer, last answer, best answer options available when grading.

bnmnetp:
Oh, I was reading the diff backward 😬 . You converted it from a reference to an integer so that will actually help a bunch.

PR #1391, opened 2019-11-26
Add: Assignment table

bjones1:
This PR implements the beta assignment table.

### DIFF
--- a/controllers/assignments.py
+++ b/controllers/assignments.py
@@ -13,6 +13,7 @@ import json
 import logging
 import datetime
 from collections import OrderedDict
+import traceback
 
 # Third-party imports
 # -------------------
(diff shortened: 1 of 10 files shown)


bjones1:
Oops, I fixed the formatting. All tests pass and this is ready.

bjones1:
I've updated this, so that it now pulls data from the db for all question types. Missing pieces that I'd like help on:

- [ ] Testing. I have good tests for mchoice, fitb, lp, and shortanswer. I need help writing tests for other problem types, since I use those rarely and am not familiar with them. The test book needs one of each question type added to it, then code added to insert correct and incorrect answers for each question type. See `tests/test_server.py` line 857.
- [ ] JS rendering. How do we show instructors the results of a parson's problem? codelens? etc. See 'static/js/assignment_table.js` line 404.

bjones1:
Note that the tests are failing because web2py.com is down.

bjones1:
The JS is now fixed and my goofy VSCode changes reverted. Remember that you'll need to reset web2py to test this out -- changes are in a module, which requires a reload of web2py, unlike models/controllers/views.

bjones1:
I've updated the PR based on the bugs you identified.

Excluding instructors is tricky. I haven't figured out how to make that work yet.

PR #1401, opened 2020-01-03
fix table name capitalization (again)

presnick:


bnmnetp:
What is the purpose of this change?

As far as I can tell its been capitalized this way since day one.  If you change it here you need to also change the table definition in models/practice.py

presnick:
I think that’s already changed, so this causes an error. It got changed
some time ago, maybe over the summer, because it was creating issues on
some Postgres installations that did or didn’t distinguish between caps and
small. I don’t remember exactly, but standardizing on lower case seemed to
be a safe resolution at the time.

bnmnetp:
If you look at the master branch on github you will see that it is still capitalized there.

bnmnetp:
Oh, I see you submitted that as a separate PR.

bnmnetp:
Attention @bjones1 -- I had to add to this PR to remove the capitalization from the table name, @presnick did this but it was not part of this PR.   HOWEVER -- web2py does not see the name change when you do a migration.  You need to manually change the table name with an alter table command:

```
alter table "user_topic_practice_Completion" rename to user_topic_practice_completion;
```

PR #1407, opened 2020-01-14
Remove sections

bnmnetp:
This removes the code for sections.  They really were not used anywhere, so most of the code was for section CRUD, and adding or removing students from sections.

@bjones1, @presnick there are a few lines in lti.py that refer to sections that I wasn't sure about just deleting. If you could take a look and let me know its ok to remove the remaining references to section in lti.py that would be great.


bnmnetp:
Is that by choice or because of your kubernetes setup?  -- couldn't you commit the databases folder to the branch on your repo you use for your production system?

PR #1411, opened 2020-02-05
LTI fixes

bjones1:
This is the PR promised from #1322, plus a few more fixes. The LTI fix involves following @presnick's [suggestion](https://github.com/RunestoneInteractive/RunestoneServer/pull/1322#issuecomment-571189115) -- the donate/pay option only applies to Academy mode. A few other items included:

1. I renamed an LTI variable from `assignment_id` to `custom_assignment_id` per LTI specs and deprecated (but still support) `assignment_id`.
2. I removed `six`, since we don't support Python 2.
3. Black fixes.
4. Comment clean-up/notes on LTI key generation.
5. Test clean-up -- added missing tables to the `TRUNCATE` list.


bnmnetp:
I'm happy to merge this, but I'll wait to make sure @presnick is ok with it.

presnick:
Sorry, but I'm not going to be able to test this on one of my servers for a
while, probably a month or more.

bnmnetp:
It seems Brian has followed your suggestions, so I think its safe to accept, which I would like to do to get the other fixes.

presnick:
OK. Sounds good. In the worst case, I'll propose some additional changes
later.

Issue #1417, opened 2020-02-25
RuenstoneServer Booting Bug


jlliu-1:
After building the RunestoneServer container in Docker, when you when you attempt to boot the respective container via docker-compose up (after setting the environment variables), the error
```
open(“/srv/web2py/logs/uwsgi.log”): No such file or directory [core/logging.c line ###]
runestone_1  | unlink(): No such file or directory [core/uwsgi.c line ####]
runestone_1  | tail: cannot open ‘/srv/web2py/logs/uwsgi.log’ for reading: No such file or directory
```
still persists, preventing the booting of the server. This works every time after having to restart Docker and the only pseudo fix I have found is executing these commands in the command prompt:

```
docker-compose up -d
CONTAINER_ID=$(echo `docker-compose ps -q runestone` |  cut -c1-12)
docker exec -it $CONTAINER_ID bash
mkdir -p /srv/web2py/logs 
touch /srv/web2py/logs/uwsgi.log
exit
docker-compose restart
```
Would there be a way to fix this bug to alleviate this additional step needed in booting process?

bnmnetp:
That is strange, I haven't run into that problem.

The tail command is the last command in the `entrypoint.sh` file and the server should have already started and created the log file by then.  

Do you have any books set up in the books folder?  Normally the server starts and then the script rebuilds the book(s) which takes a bit of time.  If not then this is probably a race condition where the server has started but the startup process isn't far enough along to have created the log yet.

jlliu-1:
Yes, I do have two books installed but the file was still never created on time and then `tail` fails.

PR #1419, opened 2020-02-26
Inserted Booting Log Bugfix #1417

jlliu-1:
Added 2 commands to prevent the missing logs error upon boot-up to fix issue #1417


bnmnetp:
I can see how this will fix #1417 but you still haven't answered the questions I asked you there.

Issue #1292, opened 2019-08-07
Filter by Student in Gradebook only matches FName


bhoffman0:
In the instructor dashboard Gradebook page, Filter by Student only matches the First Name and not the Last Name.  Also, pin the names for the gradebook.

christsona:
Hello, my name is Aaron and I can take a look at this.

PR #1423, opened 2020-03-09
Patched issue #1292

jlliu-1:
In the Grading panel (gradebook), you can currently only filter by first name. However, this pull request adds the functionality to sort by last name as well as first name.


bnmnetp:
Thanks, this is a nice update.  

I would really like to get the grade book to use the handsontable javascript library like we do for the overview reports.  That takes care of filtering and sorting on any column for us.

PR #1425, opened 2020-03-13
Copy one assignment

bnmnetp:
Per many requests, you can now copy a single assignment.


bjones1:
To do, IMHO: add a test case which shows that the current copy feature allows creating two assignments with the same name in a given course. Then add code to avoid this case.

PR #1433, opened 2020-04-02
Use webpack

bnmnetp:
This PR needs to be accepted when the webpack PR of components is accepted.  It trims the number of script and css includes dramatically


bnmnetp:
@bjones1 I think you are going to like what you see here 😄

PR #1448, opened 2020-04-30
Misc fixes

bjones1:
- In the client-side code of `admin.js`, use sid, not student's name, to select a student internally.
- Fix incorrect links in the gradebook.

### DIFF
--- a/rsmanage/setup.py
+++ b/rsmanage/setup.py
@@ -3,10 +3,10 @@ from setuptools import setup
 setup(
     name="rsmanage",
     version="0.1",
-    py_modules=["rsmanager"],
+    py_modules=["rsmanage"],
     install_requires=["Click"],
     entry_points="""
         [console_scripts]
-        rsmanage=rsmanager:cli
+        rsmanage=rsmanage:cli
     """,
 )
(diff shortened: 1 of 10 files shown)


bnmnetp:
Looking at this again, now that 5.0 is merged.

Seems like this would be a good one to have a unit test for to prevent a future regression along these lines.

bjones1:
Good point. I'll see what I can do.

bjones1:
I've updated this PR after rebasing to post-webpack master, plus added several more improvements and fixes to the tests.

rsmanage/setup.py
```diff
@@ -3,10 +3,10 @@
 setup(
     name="rsmanage",
     version="0.1",
-    py_modules=["rsmanager"],
+    py_modules=["rsmanage"],
```
bnmnetp, line 6: Wait, what?  Why change this name??

bjones1: This means we can use `python -m rsmanage` to run this module (as opposed to `python -m rsmanager`). The name change makes it match the command-line entry point `rsmanage`.

Since this is run from the server in in the `default/delete` endpoint, we need to execute this as `settings.python_interpreter -m rsmanage` and keeping the name consistent seemed best to me.

PR #1464, opened 2020-06-30
Make password optional for development/test databases

FateXii:
This is to make it easier to work on the project.

Also it follows more with postgres's allowance of optional passwords.

### DIFF
--- a/rsmanage/rsmanage.py
+++ b/rsmanage/rsmanage.py
@@ -50,8 +50,11 @@ def cli(config, verbose, if_clean):
 
     config.conf = conf
     config.dbname = re.match(r"postgres.*//.*?@.*?/(.*)", config.dburl).group(1)
-    config.dbhost = re.match(r"postgres.*//.*?@(.*?)/(.*)", config.dburl).group(1)
-    config.dbuser = re.match(r"postgres.*//(.*?):.*?@(.*?)/(.*)", config.dburl).group(1)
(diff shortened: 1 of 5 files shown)


rsmanage/rsmanage.py
```diff
@@ -51,7 +51,11 @@ def cli(config, verbose, if_clean):
     config.conf = conf
     config.dbname = re.match(r"postgres.*//.*?@.*?/(.*)", config.dburl).group(1)
     config.dbhost = re.match(r"postgres.*//.*?@(.*?)/(.*)", config.dburl).group(1)
-    config.dbuser = re.match(r"postgres.*//(.*?):.*?@(.*?)/(.*)", config.dburl).group(1)
+    if conf  != "production":
+        config.dbuser = re.match(r"postgres.*//(.*?)(:.*?)?@(.*?)/(.*)", config.dburl).group(1)
+    else:
+        config.dbuser = re.match(r"postgres.*//(.*?):(.*?)@(.*?)/(.*)", config.dburl).group(1)
```
bnmnetp, line 57: I don't really see what this accomplishes.  group(1) is still going to return the same thing in either case.  I can see that having the : inside the paren was probably wrong but not sure that changes anything in this case.

FateXii:
If no password is given group(1) is None this helps with that

bnmnetp:
Yep, I missed the additional `?` after the `)` before.

Thanks.

PR #1469, opened 2020-07-10
Automatically switch to a course if your enrolled

bnmnetp:
This handles the very annoying case that I run into a lot.

1. I'm logged in but my current course is X
2. I get a link from someone or I'm looking at an error report with a link, or I just want to quickly change to course Y
3. I previously registered for Y so it is perfectly good for me to be there
4. I don't want to stop at the change course page and have to select a course that I just clicked on a link for

This PR does the switch automatically.

I'm going to end up on the right page, but if I am registered for several courses with the same base course, I'll end up in the first one.

### DIFF
--- a/controllers/books.py
+++ b/controllers/books.py
@@ -35,6 +35,25 @@ logger.setLevel(settings.log_level)
 # -------------------------
 # None.
 
+# See `caching selects <http://web2py.com/books/default/chapter/29/06/the-database-abstraction-layer#Caching-selects>`_.
+cache_kwargs = dict(cache=(cache.ram, 3600), cacheable=True)
+
+
+def _get_current_course(auth):
+    course = (
+        db(db.courses.id == auth.user.course_id)
+        .select(
+            db.courses.course_name,
+            db.courses.base_course,
+            db.courses.allow_pairs,
+            db.courses.downloads_enabled,
+            **cache_kwargs,
+        )
+        .first()
+    )
+
+    return course
+
 
 # Supporting functions
 # ====================
@@ -62,8 +81,6 @@ def _route_book(is_published=True):
     if res and res.value == "PreTeXt":
         response.delimiters = settings.pretext_delimiters
 
-    # See `caching selects <http://web2py.com/books/default/chapter/29/06/the-database-abstraction-layer#Caching-selects>`_.
-    cache_kwargs = dict(cache=(cache.ram, 3600), cacheable=True)
     allow_pairs = "false"
     downloads_enabled = "false"
     # Find the course to access.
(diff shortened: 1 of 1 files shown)


controllers/books.py
```diff
@@ -74,25 +91,41 @@ def _route_book(is_published=True):
...
+                    & (db.courses.base_course == base_course)
+                )
+                .select()
+                .first()
             )
-            redirect(URL(c="default", f="courses"))
+            if res:
+                # change user to this course
+                db(db.auth_user.id == auth.user.id).update(course_id=res.courses.id)
+                db(db.auth_user.id == auth.user.id).update(
```
bjones1, line 114: Can't you combine these two updates into one statement?

controllers/books.py
```diff
@@ -74,25 +91,41 @@ def _route_book(is_published=True):
...
+                    course_name=res.courses.course_name
+                )
+                auth.user.update(course_name=res.courses.course_name)
+                auth.user.update(course_id=res.courses.id)
```
bjones1, line 118: Same.

controllers/books.py
```diff
@@ -74,25 +91,41 @@ def _route_book(is_published=True):
...
+                current_course = res.courses.course_name
+                course = _get_current_course(auth)
+                session.flash = (
+                    f"Automatically changing your current course to {current_course}"
```
bjones1, line 122: I'm still on Python 3.6 -- would you mind using `format` instead?

controllers/books.py
```diff
@@ -245,7 +278,7 @@ def dummy():
         questions = _exercises(base_course, chapter)
     logger.debug("QUESTIONS = {} {}".format(subchapter, questions))
     return dict(
-        course_name=course.course_name,
+        course_name=current_course,
```
bjones1, line 281: I don't think this is assigned for all cases.

controllers/books.py
bnmnetp: I thought we agreed that we could move on....  I've been freely coding with f strings in the components as well.

3.9 is in beta and will be released in October.  I think should support the current release and 1 release back and be testing on the current beta so we are ready to support the current release.

controllers/books.py
bnmnetp: I don't know what you mean?

I am interested in what you think of the concept of this PR.

bjones1: I misunderstood -- was giving more of a code review. I like the concept.

I meant, does `current_course` get assigned for all cases? Glancing through the code, I only see its being assigned inside an `if` statement, but I might have missed something.

controllers/books.py
bnmnetp: regardless, I just tried and at least in 3.6.9 f-strings seem to work fine.

bjones1: Oops, you're right. F-strings are OK (supported in Python 3.6).

How do you keep your server's Python current? I'd been using whatever ships with the LTS version of Ubuntu, and I'm hesitant to upgrade that since Python touches so many pieces of the OS.

controllers/books.py
bnmnetp: I think it is good.

if course is not found then we set it.
but if course is not found then we are either going to set it or redirect the user back to the change course page.

controllers/books.py
bnmnetp: updated

controllers/books.py
bnmnetp: I use the ppa provider for python and use apt.  On my dev server I use pyenv so I can quickly change around.  No problems in any cases.

controllers/books.py
bnmnetp: Except for the case where the user is not logged in! 
Caught by the unit tests too!

bjones1: +1 for unit tests!

controllers/books.py
bjones1: Great, thanks! I'll try that out soon. Right now, students are taking a test on my server, so I'll have to wait for some downtime...

Issue #1470, opened 2020-07-11
Error handling in questions_report


bnmnetp:
I'm getting the following crash. when trying to view the assignment table

```
Traceback (most recent call last):
  File "/home/bmiller/Runestone/web2py/applications/runestone/controllers/assignments.py", line 1112, in grades_report
    auth.user.course_name, request.vars.chap_or_assign,
  File "/home/bmiller/Runestone/web2py/applications/runestone/modules/questions_report.py", line 585, in query_assignment
    None,
KeyError: 'epearson'
```
The `epearson` in question should be a student in the course.

I think this brings up two potential improvements.

1. make sure grades contains a key before we use it.
2. in the populateAssignmentTable function we should check for data.errors before we iterate.  This way we can report to the user that something bad happened and ask them to report it rather than leaving them hanging with Loading...

bjones1:
Will do. Is there any way I can recreate this? Or would you allow me to either run queries on your db, or run queries for me? I'd like to understand how this user didn't get found by the two queries in `_headers_query` and how to fix that.

bnmnetp:
I'm happy to run the queries for you, even be on zoom and running so we can discuss in real time while doing so.

bjones1:
Wonderful! I'll put together some notes then find a time. Thanks!

bjones1:
When are you free to do some debug?

PR #1472, opened 2020-07-13
Fix: Provide an error report when the assignment table raises an exception

bjones1:
Better error reporting per #1470.

### DIFF
--- a/modules/questions_report.py
+++ b/modules/questions_report.py
@@ -271,15 +271,6 @@ def _headers_query(
 
     # **user_id headings query**
     ## ---------------------------
-    # The base query for this entire function. It's used to get information about each user_id/div_id combination.
-    query = (
-        # Choose seleted questions.
-        query_questions
-        # Join them to ``useinfo``.
-        & (db.questions.name == db.useinfo.div_id)
-        # Select only questions in the provided course.
-        & (db.useinfo.course_id == course_name)
-    )
     # Get information about each user_id.
     select_args = [db.auth_user.first_name, db.auth_user.last_name, db.auth_user.email]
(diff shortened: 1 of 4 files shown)


bjones1:
I just pushed an update with a fix for my guess at what's wrong. Let me know when you're available to test on your server.

bjones1:
Perhaps +1 for unit tests again. I created a unit test for the case I think my code missed, verified that the tests fails (with the same sort of exception as you saw), then verified the new code makes it pass. Let me know if this works.

bnmnetp:
Yikes, this seemed to help on my dev machine, but on the live machine it is bringing Postgres to its knees!

bjones1:
Oh no! I've used some nice tools to look at postgres performannce -- let me know if you want help/info on those. There's some comments on them in `questions_report.py`. (I needed these for another type query, not for what we're using right now.)

bnmnetp:
I just reverted this merge as the query seems to run whenever you select an assignment on the grading page.  

The query starts and that cpu goes immediately to 100% for Postgres.  I had all of the cpus on academy pegged at 100% until I could manually kill off these queries, some of which were running for more than 10 minutes.

bjones1:
Holy indexes, batman! Let me know how I can help...

bnmnetp:
Not even sure where to start... DISTINCT is always a warning flag for performance.  Unfortunately I don't really have a lot of time for this right now, as the exam generator project has to be my top priority.  It would be nice to at least get the js error handling live so people are not left wondering.

bjones1:
The first commit in this PR fixes that. Do you want that as a separate PR? I might also work on an alternative approach and see if that helps.

PR #1478, opened 2020-07-15
Fix: Use Sphinx section numbers, not Runestone chapter/subchapter numbers.

bjones1:
This should be merged after https://github.com/RunestoneInteractive/RunestoneComponents/pull/1017. It uses Sphinx section numbers which that PR assigns.

I didn't update the use of `chapter_num` and `sub_chatper_num` in `dashboard.py`, a place I fear to tread...

### DIFF
--- a/controllers/books.py
+++ b/controllers/books.py
@@ -267,7 +267,6 @@ def _subchaptoc(course, chap):
         & (db.chapters.course_id == course)
         & (db.chapters.chapter_label == chap)
     ).select(
-        db.chapters.chapter_label,
         db.sub_chapters.sub_chapter_label,
         db.sub_chapters.sub_chapter_name,
         orderby=db.sub_chapters.sub_chapter_num,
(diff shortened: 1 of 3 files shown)


controllers/books.py
```diff
@@ -267,9 +267,6 @@ def _subchaptoc(course, chap):
         & (db.chapters.course_id == course)
         & (db.chapters.chapter_label == chap)
     ).select(
-        db.chapters.chapter_num,
```
bnmnetp, line 270: These are the lines that take away the numbers from the "This Chapter" menu.

If we can populate the chapter and subchapter table more accurately then we should do it.  Or if you have a different mechanism in mind to provide numbers here?

bjones1:
Sphinx already numbers chapters/subchapters if the toctree includes `:numbered:`, so we shouldn't duplicate that.

bnmnetp:
Merged with my changes.  Thanks!

PR #1485, opened 2020-07-30
provide endpoint for lti store

bnmnetp:
* Allows TC to select an assignment !

Problems

 - [ ] the test tool keeps telling me that I have a mismatch on the signature.  My assumption is that I would just echo back the signature I got on the request.  But maybe that is a bad assumption.

### DIFF
--- a/controllers/books.py
+++ b/controllers/books.py
@@ -167,7 +167,7 @@ def _route_book(is_published=True):
         raise HTTP(404)
 
     # See if this is static content. By default, the Sphinx static directory names are ``_static`` and ``_images``.
-    if request.args(1) in ["_static", "_images"]:
+    if request.args(1) in ["_static", "_images"] or book_path.endswith(("css","png","jpg")):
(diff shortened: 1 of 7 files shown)


bnmnetp:
@bjones1 Can you take a look at this?  I think I've got what we need in place for selecting an assignment according to the spec.  But I'm having problems with the OAuth signature not matching.  I may just be missing something obvious that a second pair of eyes will catch.

bjones1:
Just a wild guess -- did you see this from `lti.py`?

``` Python
        # Fix encoding -- the signed keys are in bytes, but the oauth2 Request constructor translates everything to a string. Therefore, they never compare as equal. ???
        if isinstance(oauth_request.get("oauth_signature"), str):
            oauth_request["oauth_signature"] = oauth_request["oauth_signature"].encode(
                "utf-8"
            )
```

bnmnetp:
Thanks, that did not hellp.

I'm pretty sure that I am either including some element into the signature calculation that I should not be, or that I am missing something from the signature calculation.

PR #1513, opened 2020-09-09
Progress course only

bnmnetp:
This PR fixes a longstanding annoyance that the chapter progress buttons on the pages as well as the icons in the table of contents will reflect ANY book with that sub_chapter_label.  So a student could start a course in the basecourse and then switch to their instructors course and it would look like they had completed the chapter in the instructors course.  But they would not get credit because the work was done in another course.

This makes the buttons and the icons only reflect work done in the current course so it is consistent with the grader.

### DIFF
--- a/controllers/default.py
+++ b/controllers/default.py
@@ -212,16 +212,16 @@ def index():
                 db(
                     (db.user_sub_chapter_progress.user_id == auth.user.id)
                     & (db.user_sub_chapter_progress.chapter_id == chapter_label)
+                    & (db.user_sub_chapter_progress.course_name == course.course_name)
                 ).count()
                 == 0
             ):
                 db.executesql(
-                    """
-                    INSERT INTO user_sub_chapter_progress(user_id, chapter_id,sub_chapter_id, status, start_date)
-                    SELECT %s, chapters.chapter_label, sub_chapters.sub_chapter_label, -1, now()
-                    FROM chapters, sub_chapters where sub_chapters.chapter_id = chapters.id and chapters.course_id = '%s';
+                    f"""
+                    INSERT INTO user_sub_chapter_progress(user_id, chapter_id,sub_chapter_id, status, start_date, course_name)
+                    SELECT {auth.user.id}, chapters.chapter_label, sub_chapters.sub_chapter_label, -1, now(), '{course.course_name}'
+                    FROM chapters, sub_chapters where sub_chapters.chapter_id = chapters.id and chapters.course_id = '{course.base_course}';
                 """
-                    % (auth.user.id, course.base_course)
                 )
         except Exception as e:
             logger.error(f"Select Course got Error {e}")
--- a/controllers/default.py
+++ b/controllers/default.py
@@ -556,6 +556,10 @@ def privacy():
     return dict(private={})
 
 
+def wisp():
+    return dict(wisp={})
+
+
 def ct_addendum():
     return dict(private={})
 
--- a/controllers/default.py
+++ b/controllers/default.py
@@ -217,11 +217,16 @@ def index():
                 == 0
             ):
                 db.executesql(
-                    f"""
+                    """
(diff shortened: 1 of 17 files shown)


controllers/default.py
```diff
@@ -212,16 +212,16 @@ def index():
...
+                    & (db.user_sub_chapter_progress.course_name == course.course_name)
                 ).count()
                 == 0
             ):
                 db.executesql(
-                    """
-                    INSERT INTO user_sub_chapter_progress(user_id, chapter_id,sub_chapter_id, status, start_date)
-                    SELECT %s, chapters.chapter_label, sub_chapters.sub_chapter_label, -1, now()
-                    FROM chapters, sub_chapters where sub_chapters.chapter_id = chapters.id and chapters.course_id = '%s';
+                    f"""
```
bjones1, line 220: The change here is mainly using f-strings instead of the % operator, but the whole things scares me from a security standpoint. I know it's more pain, but I think **all** raw SQL with replacement strings should either be written using DAL or using the [placeholders argument of executsql](http://www.web2py.com/books/default/chapter/29/06/the-database-abstraction-layer#executesql).

tests/test_ajax2.py
```diff
@@ -813,6 +813,10 @@ def test_updatelastpage(test_client, test_user_1, runestone_db_tools):
         db(
             (db.user_sub_chapter_progress.user_id == test_user_1.user_id)
             & (db.user_sub_chapter_progress.sub_chapter_id == "subchapter_a")
+            & (
```
bjones1, line 816: It would be nice to have a test case showing before and after behavior -- a test that fails before this fix, since progress is marked on two different courses which share the same base course, vs. no failure afteward.

controllers/default.py
bnmnetp: I agree with you in general.  But in this case there is no form to process, the id field is assigned by the system. course name and base course name also come from the database and are inserted with DAL.  This is one of those odd queries that I don't have any idea how to do in DAL, and the readability of this is just so much nicer.

bjones1: True -- DAL doesn't support this. Perhaps

``` Python
db.executesql("""
    INSERT INTO user_sub_chapter_progress(user_id, chapter_id,sub_chapter_id, status, start_date, course_name)
    SELECT %s, chapters.chapter_label, sub_chapters.sub_chapter_label, -1, now(), %s
    FROM chapters, sub_chapters where sub_chapters.chapter_id = chapters.id and chapters.course_id = %s;
""", placeholders=(auth.user.id, course.course_name, course.base_course))
```

tests/test_ajax2.py
```diff
@@ -533,6 +533,37 @@ def test_GetLastPage(test_client, test_user_1):
     assert res[0]["lastPageChapter"] == "Test chapter 1"
 
 
+def test_LastPageCrossCourse(test_client, test_user_1, runestone_db_tools):
```
bjones1, line 536: Nice!

controllers/default.py
bnmnetp: I know it safer but I think that syntax is mean and misleading for anyone who knows printf.  Why is everything regardless of type %s?  It makes no sense.  If you are going to do that why not go all the way and use %d for ints and %f for floats.

bjones1: I agree. This syntax is just terrible. However, it's the only way I know to sanitize SQL queries. Sigh. Looks like a place for some good docs about awful syntax. Very ugly...

bnmnetp: slightly more readable would be the `%(userid)s` route and then you can pass a dictionary of placeholders.

bjones1: Good idea, I like that!

bnmnetp: Looks like only 20 places in the code where we use db.executesql -- I'm going to make sure they are all sanitized. Even the ones I think are perfectly safe. 😬

bjones1: Awesome. I'll definitely sleep better at night. "After all, what could possibly go wrong..."

PR #1533, opened 2020-09-23
acerror_log is redundant - remove all access but not the table just yet.

bnmnetp:
The only question is whether we should save anonymous users' data to code?  The only real issue is whether it has research value, and at this point I don't think it really does.


bjones1:
This looks fine to me. As you say, I don't see a lot of value for this data.

I'd suggest adding a TODO to the appropriate file in models/ with a note that it's no longer used as of commit xxx and should be removed at some point in the future.

PR #1544, opened 2020-09-30
modified percentage calculation that appears on progress page

spontarelliam:
modified pct calculation to be scaled from 0 to 100 and converted to an int to drop excess precision.


bnmnetp:
Thanks, this is better.

PR #1554, opened 2020-10-24
(2) Repopulate Time Limit and Feedback checkboxes correctly, (3) Add Preview assignment button, (4) Fix formatting issue on installation document

vqum:
Invert default of Provide Feedback to No Feedback - Require instructor to check the Feedback box in order for students to receive feedback on timed assignments because in the assessment setting, accidentally neglecting the Feedback option in assignment setup and defaulting to provide feedback could pose an academic integrity issue. This change is in coordination with a change in RunestoneComponents:
https://github.com/RunestoneInteractive/RunestoneComponents/pull/1088

Repopulate Time Limit and Feedback checkboxes correctly - previously input and saved values for Time Limit and Feedback were not repopulating no revisiting the assignment administration

Add Preview assignment button - add button to assignment administration that opens the assignment in a separate tab for the instructor to view it before having to make it visible to students

### DIFF
--- a/controllers/admin.py
+++ b/controllers/admin.py
@@ -1848,6 +1848,7 @@ def get_assignment():
     assignment_data["description"] = assignment_row.description
     assignment_data["visible"] = assignment_row.visible
     assignment_data["is_timed"] = assignment_row.is_timed
+    assignment_data["time_limit"] = assignment_row.time_limit
     assignment_data["from_source"] = assignment_row.from_source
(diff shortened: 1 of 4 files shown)


vqum:
Retracting invert default of Provide Feedback to No Feedback - from Brad: "it is changing default behavior (not backward compatible) of loads of existing timed exams."

bnmnetp:
Merged.  Thanks.

PR #1578, opened 2021-01-08
Update jobeinabox image to use files for unit tests

nitramsivart:
The current state of the jobe server in the RunestoneServer repository does not reflect the true status of the code that is hosting runestone.interactive. When following the directions to host a server via Docker, code in CSAwesome (and possibly other books) does not compile, because the jobe server is missing some necessary files, such as junit.jar.

See also this related question (which prompted some helpful private discussion with Brad): https://groups.google.com/g/runestone_instructors/c/3_2hElGuiYk

Note that https://github.com/RunestoneInteractive/jobe *does* have the necessary updates, but updates need to be made to use the proper image (and extra files) when launching the images with Docker Compose.

After the changes in this PR, I was able to get CSAwesome working on my local server.

Questions for reviewer:

1. Does this seem like the right places to modify code? Alternatives I considered included (a) modify https://github.com/RunestoneInteractive/jobe itself & hardcode the necessary .java / .jar files (I avoided this because it requires changing two repos if a book needs e.g. a newer .class file) or (b) fork a new runestone/jobeinabox repo (IMO it's better to just build a docker image based on the old image, instead of forking a whole repo. This reduces the pain of staying in sync, though it does have a higher risk of being broken by some random docker image update).
2. Does it seem reasonable to include the pre-included .class files? Or the .jar files, for that matter. I could alternatively compile / build these as part of the Docker Build process.
3. Is there a better way to do the docker extension I'm doing here?
4. Any other general suggestions?


bnmnetp:
This is great.  I like this because it keeps the changes contained to the RunestoneServer repo while allowing us to customize JOBE.

I think it is fine for us to include these .jar files as it keeps us in control over when to upgrade them etc.  I'm not sure the .class files need to be included as they would get built automatically. (I think)

I'm no Docker expert, my first thought was to Fork the jobeinabox repo, but I like this better.

I was wondering about the use of the volume versus running mkdir and copying in the Dockerfile?  Any advantage to simply copying instead of using the volume?  I guess for updating it means we don't have to rebuild the container.

Now that I have this I am reminded that I have a couple of header files that need to be added for C++ unit testing.  But that is easy to see how to do that now that you have the ball rolling.

bnmnetp (approved):

nitramsivart:
You're right, the .class files are indeed built automatically. I was originally getting an error, but I now realize that this was simply because the jobe server did not have permissions to write to the `files/` directory -- I've double the checked that the permissions aren't an issue when everything has been freshly downloaded from GitHub (testing this now).

My main intention with the volume instead of copying was to avoid needing to rebuild the container. But, I'd be happy to switch it to copying, if that seemed better.

Otherwise, this should be good to merge!

PR #1584, opened 2021-01-15
Feature/lti docs

Venator-Fox:


bnmnetp:
Did your Fork have a `docs/lti.rst` file?  This should really go in in that file in its own section rather than a separate file.  The master branch does have that file as I just accepted a PR to fix a couple typos today.

bnmnetp (changes requested):
This is great, thanks.

Could you please update index.rst (in root directory) to reference this file so it is included in the table of contents?

Have you tried moodle with runestone.academy?   I'm assuming that it should work with runestone.academy if you have it working with your own instance of the Runestone Server.

Thanks -- really do appreciate PRs!

bnmnetp (approved):

PR #1592, opened 2021-02-02
Fix: LTI course_id was a string, instead of an int.

bjones1:


controllers/lti.py
```diff
@@ -46,11 +47,15 @@ def index():
     message_type = request.vars.get("lti_message_type")
     course_id = _param_converter(request.vars.get("custom_course_id", None))
 
-    if course_id and not course_id.isnumeric():
-        course_id = (
-            db(db.courses.course_name == course_id).select(**SELECT_CACHE).first()
-        )
-        course_id = course_id.id
+    if course_id:
```
bnmnetp, line 50: Just curious, what was causing the failure to change it to this?

bjones1: f9bb8ae4 now uses the course_id, which was actually incorrect all along in the LTI code (stored as a string instead of an int), which caused LTI failures when accessing the admin interface.

PR #1595, opened 2021-02-05
LTI docs update

bjones1:


bnmnetp:
The `lti_selfhost` file was just contributed for people who are running their own servers.  Can you explain why you just deleted it?

bjones1:
Most of the file is about LTI with Moodle -- only the first bullet item (reproduced below) was about self-hosting. So, I combined these two, and eliminated sections that made no sense (self-generating keys).

```
#.  Install Runestone Server

    #.  If you would like your installation to be LTI only, set `settings.lti_only_mode = True` in `models/0.py`
```

PR #1604, opened 2021-02-17
[Draft PR] Select question :toggle: functionality for debugging active code "Save & Run" button

vqum:
In conjunction with changes to RunestoneComponents, see https://github.com/RunestoneInteractive/RunestoneComponents/pull/1140


bnmnetp (commented):
Are any of these are really necessary? get_question_source will already get the previously selected question, and things seemed to work fine for me even without applying this PR.  What am I missing?

vqum:
in `admin.py`, the `@auth.requires` change is so that student accounts can access the `htmlsrc()` function for toggle to work. `htmlsrc()` is accessed in toggle during initial loading to populate the dropdown menu with each question's question type, when previewing the question to render it in the preview div, and when switching the question to render it in the question div

in `ajax.py` in `get_question_source`
`if toggle:
        prev_selection = (
            db(
                (db.selected_questions.sid == auth.user.username)
                & (db.selected_questions.selector_id == selector_id)
            )
            .select()
            .first()
        )
        if prev_selection:
            questionid = prev_selection.selected_id
        else:
            questionid = request.vars["questions"].split(",")[0]`
This condition checks for whether the `:toggle:` flag is set in the selectquestion, and if it is set then it will render the first question in the selectquestion's `:fromid:` list as opposed to picking one at random as is the default functionality of `get_question_source`

controllers/ajax.py
```diff
@@ -1726,6 +1727,24 @@ def get_question_source():
...
+                & (db.selected_questions.selector_id == selector_id)
+            )
+            .select()
+            .first()
+        )
+        if prev_selection:
+            questionid = prev_selection.selected_id
+        else:
+            questionid = request.vars["questions"].split(",")[0]
+    else:
```
bnmnetp, line 1744: This else is wrong.  the `if toggle` is at the outermost level of the function and this is keeping the rest of the function from executing

bnmnetp: It seems like there must have been some other error checking going on?  I just commented out the else and all of our current unit tests for get_question_source do pass.

PR #1621, opened 2021-03-18
Shared tests

bjones1:
This provides server-side testing of the poll component by itself and when inside a selectquestion.


requirements-dev.txt
```diff
@@ -4,355 +4,125 @@
...
+pyflakes==2.2.0           # via flake8
+pygments==2.7.4           # via codechat, runestone, sphinx
+pylint==2.6.0             # via -r requirements.in
+pyparsing==2.4.7          # via packaging, runestone
+pyrsistent==0.17.3        # via jsonschema
+pytest==6.2.2             # via -r requirements-dev.in
+python-dateutil==2.8.1    # via -r requirements.in, pandas
+pytz==2021.1              # via -r requirements.in, babel, celery, pandas, runestone
+pyvirtualdisplay==2.0     # via -r requirements-dev.in
+pywin32==300              # via locust
```
bnmnetp, line 92: pywin32 does not install on mac or windows....

bnmnetp: Or maybe we should not `pip-compile` on windows ??

bjones1: Grrr. It looks like locust brings this in unconditionally on Windows. I added an environment marker to try and work around it. Hopefully, that will work.

bjones1:
Sounds great! Thanks.

bnmnetp:
OK, I just released 5.5.17 but the requirements-dev.txt is asking for 5.5.12

bjones1:
Hmm, I realized that testing against runestone from PyPI won't work, since we need the server to have access to all the `test/` directories. These directories aren't included in the standard runestone package (for good reason). How about installing and testing against the latest version of the components from Github? If not, we can always specify a specific github tag/release to build.

bjones1:
@bnmnetp, this is ready. I used Runestone Components directly from Github master, instead of a released version.

bnmnetp (approved):

PR #1624, opened 2021-03-23
Add total points to bottom of questions list

vqum:
Duplicate the total points that is near the top of the instructors' assignments admin page and places this copy at the bottom of the questions list. This is so that changes to the total points of an assignment can be more easily seen while making changes to an assignment's question list without needing to scroll back up to the top of the page.
image

See https://github.com/RunestoneInteractive/RunestoneComponents/issues/1143

### DIFF
--- a/views/admin/assignments.html
+++ b/views/admin/assignments.html
@@ -645,17 +645,6 @@ $(function () {
         },
     });
 });
-  
-var totalPointsDiv = document.getElementById('totalPoints');
-var totalPointsCopy = document.getElementById('totalPointsCopy');
-var observer = new MutationObserver(function() {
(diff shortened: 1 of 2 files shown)


views/admin/assignments.html
```diff
@@ -644,6 +645,17 @@ <h5 style="text-align: center">Search Question Bank</h5>
         },
     });
 });
+  
+var totalPointsDiv = document.getElementById('totalPoints');
```
bnmnetp, line 649: I like this, but I would rather have this added to admin.js.

I'm slowly trying to get the javascript out of the html files and into js files to make it easier to maintain.

Issue #1613, opened 2021-03-10
I would like the ability to copy an assigngment from the same course


barbarer:
**Is your feature request related to a problem? Please describe.**
I have to create two assignments for every midterm - one for people with the normal amount of time and one for people with extending time.  It would save me time if I could copy the assignment but this is currently limited to copying from another course and not the same course.

**Describe the solution you'd like**
Allow copying from the same course - you could automatically name it CopyOfOrigName.

**Describe alternatives you've considered**
I currently have to create a second assignment with the same questions in the same order and same grading.

**Additional context**
I use this for midterms.

vqum:
For visibility to avoid duplication of effort, I'm taking a shot at building this feature this week.

PR #1629, opened 2021-03-31
Duplicate an assignment

vqum:
Addressing https://github.com/RunestoneInteractive/RunestoneServer/issues/1613 "I would like the ability to copy an assignment from the same course"

**Changes made to:**
assignments.html 
- Added a form group to the 'Add' modal window to provide the option of creating a new assignment that copies the settings, readings, and questions from an existing assignment
image

admin.js
- Added function `duplicateAssignmentsList` that updates the Duplicate From dropdown list to match the list of existing assignments
- Updated function `createAssignment` with a condition to check whether the newly created assignment is intended to be a copy, and if so then retrieve the data of the source assignment and set the data of the newly created assignment to match.

----------------------------------------------------------------------------------------------

Known bugs
- Copied readings and questions lists do not preserve the same order. I was experimenting with adding `async` and `await` to various places within admin.js `createAssignment` to try to preserve the order, but all of my attempts result in a blank newly created assignment with nothing copied at all.
- Sometimes individual readings or questions fail to be copied. Unknown cause.
- Sometimes the total points of the newly created assignment doesn't match that of the source assignment. Upon adding another question to the new assignment then removing that question, the total points updates itself to match correctly.
image

### DIFF
--- a/controllers/admin.py
+++ b/controllers/admin.py
@@ -1003,25 +1003,21 @@ def createAssignment():
                 points=old_assignment.points,
                 threshold_pct=old_assignment.threshold_pct,
                 duedate=old_assignment.duedate,
-                released=old_assignment.released,
                 allow_self_autograde=old_assignment.allow_self_autograde,
                 visible=old_assignment.visible,
                 is_timed=old_assignment.is_timed,
                 time_limit=old_assignment.time_limit,
-                from_source=old_assignment.from_source,
                 nofeedback=old_assignment.nofeedback,
                 nopause=old_assignment.nopause,
                 description=old_assignment.description,
             )
-            old_questions = db(
-                db.assignment_questions.assignment_id == old_assignment.id
-            ).select()
+            old_questions = db(db.assignment_questions.assignment_id == old_assignment.id).select()
             for q in old_questions:
                 dq = q.as_dict()
-                print(dq)
                 dq["assignment_id"] = newassignID
                 del dq["id"]
                 db.assignment_questions.insert(**dq)
+            db.commit()
         except Exception as ex:
             logger.error("ERROR DUPLICATING ASSIGNMENT", ex)
             return json.dumps("ERROR")
--- a/controllers/admin.py
+++ b/controllers/admin.py
@@ -105,7 +105,8 @@ def doc():
 
 
 @auth.requires(
-    lambda: verifyInstructorStatus(auth.user.course_id, auth.user), requires_login=True,
+    lambda: verifyInstructorStatus(auth.user.course_id, auth.user),
+    requires_login=True,
 )
 def assignments():
     """
@@ -151,7 +152,8 @@ def assignments():
 
 
 @auth.requires(
-    lambda: verifyInstructorStatus(auth.user.course_id, auth.user), requires_login=True,
+    lambda: verifyInstructorStatus(auth.user.course_id, auth.user),
+    requires_login=True,
 )
 def practice():
     response.title = "Practice"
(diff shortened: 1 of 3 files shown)


bnmnetp:
Have you checked to be sure the sorting_priority values in the assignment_questions table are copied correctly?  That is what determines the order.

bnmnetp:
I think you could simplify this and just make use of the `admin/copy_assignment` api call then the ordering and the points are all taken care of without the hassle of javascript async code.

vqum:
I wasn't aware that there was an existing `copy_assignment` in `admin.py`. I'll look into that, thanks.

vqum:
Moved the duplication functions from client-side in admin.js to server-side in admin.py `createAssignment`.

Question:
What are `threshold_pct`, `released`, `allow_self_autograde`, and `from_source`? I saw them in the assignment data, so I wrote it such that the new assignment would copy those values from the original assignment as well, but I don't know what these do nor whether it's necessary to copy these.

bnmnetp:
`threshold_pct` is for reading assignments and is the number of actions on the page a student must complete to get the points for that page.
`released` is that flag that says its OK to show students the grades... That should start out as false so you probably don't want to copy that.
`allow_self_autograde` lets the students run the autograder for themselves from the assignment page.  Should be copied.
`from_source` means this assignment was created in an RST file as part of a book rather than from the web interface.  So I think this should be false rather than copied.

bnmnetp:
We use `black` to reformat our code in a consistent way.  I have my editor set up to automatically run it whenever I save.  The tests are failing for this PR because of a code formatting issue.

vqum:
It's still failing `black` even though I ran it.
image

Is there something else that I'm missing?

bnmnetp:
If you run `black --version` are you running 20.8b1 ?

vqum:
Strange that `pip install`ed an older version

bnmnetp:
The `requirements-dev.txt` file has all of the versions pinned so you can just use that to make sure you have the same dev environment as I do and have the exact same environment that the tests run in.

bnmnetp:
Thanks for your work on this Vincent.  One more request before I go ahead and merge.

Could you take a look at test_server.py and see if you can add a unit test for createAssignment that tests this new functionality?

vqum:
I can take a shot at writing a unit test for this - at first glance, it seems I could add to the existing `def test_assignments()`, or would you rather it be a separate test?

However, I've never gotten Runestone Server tests to run properly on my machine, so I'm not sure how to verify what I write.

bjones1:
What OS do you use? I run my Runestone Server tests under Windows, and it works well. I don't run in a Docker container.

bnmnetp:
If you have a working server in a Docker container (which I believe you do!??) you can use the script in the `Scripts` folder of the Runestone server called `dtest` to start and run the unit tests in the docker container.

I think you are right that it may be easier to modify test_assignments.

vqum:
I am on Windows and I do not have a Docker container. Interestingly, it seems to have ran properly testing just now, so I suppose I'll run with it.

image

vqum:
Would something this simple be sufficient:
`# Duplicate an assignment -- using createAssignment`
`test_client.post("admin/createAssignment", data=dict(name=name_4, duplicate=assign1.id))`
`assign4 = (`
`    db(`
`        (db.assignments.name == name_4)`
`        & (db.assignments.course == test_instructor_1.course.course_id)`
`    )`
`    .select()`
`    .first()`
`)`
`assert assign4`

If it were to test copying the settings and the questions as well, then I think it would need something similar to the question creation lines within `test_grades_1`, and I don't want to overcomplicate it unnecessarily.

bnmnetp:
You could do one more query and just make sure that the number of questions matches the number of questions you are copying from without comparing all of the fields.  I think that would be just fine.

Issue #1628, opened 2021-03-23
Sort the assignment alphabetically in the drop down list on the assignments page in the instructor's interface.


barbarer:
The list of assignments doesn't seem to be in any particular order on the assignments page in the instructor's interface.  This makes it hard to find an assignment.  I would like them to be listed in alphabetic order by assignment name.

PR #1631, opened 2021-04-07
Sort assignments list alphabetically in instructor's admin interface (quick fix temporary solution)

vqum:
Addressing https://github.com/RunestoneInteractive/RunestoneServer/issues/1628 Sort the assignment alphabetically in the drop down list on the assignments page in the instructor's interface

Reorder the options in the assignment list select dropdown alphabetically while preserving their assignment ID and data. This is only meant to be a temporary quick fix that will be replaced by a better longer term solution.
image

### DIFF
--- a/static/js/admin.js
+++ b/static/js/admin.js
@@ -1293,6 +1293,18 @@ function getAssignmentId() {
     return assignlist.options[assignlist.selectedIndex].value;
 }
 
+function assignListAlpha(assignlist) {
+    var currentAssignment = assignlist.value;
+    var options = $(assignlist).children().toArray();
+    options.sort(function(a,b) {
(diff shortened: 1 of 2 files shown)


static/js/admin.js
```diff
@@ -1293,6 +1293,18 @@ function getAssignmentId() {
     return assignlist.options[assignlist.selectedIndex].value;
 }
 
+function assignListAlpha(assignlist) {
```
bnmnetp, line 1296: At first I thought, why not just have the database sort these for us.  But I appreciate what you did here is better as it results in a case insensitive sort of the assignment names. -- definitely better for this purpose.

bnmnetp (approved):

PR #1636, opened 2021-04-12
[Draft PR] Assignments list sorting

vqum:
Addressing https://github.com/RunestoneInteractive/RunestoneComponents/issues/1156 "Sort Assignments in student assignment page from the table of contents in reverse order (most recent first)"
and https://github.com/RunestoneInteractive/RunestoneComponents/issues/1144 "students would like some improvements to the assignments page in the table of contents"

Wanted to add column sorting functionality to the assignments list so that students have ability to order the list according to more attributes - mouseover of column header changes cursor to pointer to indicate click interaction, clicking reverses the column sort order
image
image
image

----------------------------------------------------------------------------------------------

Questions:
- I'm looking to add a few more columns to the table based on assignment data attributes. Does each assignment have any data attributes indicating whether the student has already completed it, and how many points the student received for it?
- Is this a good approach to try to address these issues? Since I'm also considering sortable tables to address other issues (https://github.com/RunestoneInteractive/RunestoneServer/issues/1628#issuecomment-809621574), should I try to generalize the functions to be usable across multiple pages?
- I know we want to separate JavaScript from the HTML - to which .js file might the code be moved to? It didn't seem to me that assignments.html currently references any .js files.


bnmnetp:
Good Questions @vqum and thanks for taking this on.  Here are my thoughts on your questions

We have used the handsontable javascript library for some tables, and I am hoping for me or someone to rewrite the gradebook table with it as well.  Right now it is used in a few places including the Chapter Progress reports.  I'm 100% in favor of standardizing on SOME package for doing tables.  Handsontable seems reasonably easy to use and gives you sorting and filtering on columns for free.

We do not really have a "submit assignment" button for runestone.  We just grade whatever the student has done by the deadline.  A students score is recorded in the `grades` table with a foreign key to the `assignments` and `auth_user` table.  I think the `grades` table is the best place to add columns with student specific state. Adding a column called submitted that would allow a student to mark an assignment as ready for grading would be great.  I'm not sure we **want** to display their score unless the instructor has released the grades for that assignment, (indicated in the assignments table released column) so maybe adding that submitted column (as a datetime type) would be a good idea and if released we can show the score there.

So many functions have been dumped into admin.js. it would be great to see admin.js refactored  into a few files like:

* grading.js
* assignments.js
* practice.js
* dashboard.js
* admin.js -- for common stuff

Assignments.html is a bit of a mess to be honest and I would love to see the javascript moved out of there into assignments.js

PR #1645, opened 2021-04-15
Duplicate assignment testing

vqum:
I'm receiving an error when I try to compare the duplicated assignment to the source. I'm checking the assignment points value as a proxy for number of questions, and I think the total points should be 1 + 2 + 3 + 4 = 10, but neither assignment seems to have this value. Can you see what could be causing this?
image


bnmnetp:
Really hard to guess from that little snippet.

If you run the tests with the -pdb option it will drop you into the python debugger and you can then look at the variables, but more importantly, you will be able to query the test database yourself in the state it was in when the debugger started.

PR #1646, opened 2021-04-18
Added error message for existing student users

Eaw7:
In response to issue 1594 about including two error cases for existing usernames, I have added the following code to determine if the student already exists and to add an error message for this case.

### DIFF
--- a/controllers/admin.py
+++ b/controllers/admin.py
@@ -2485,11 +2485,11 @@ def _validateUser(username, password, fname, lname, email, course_name, line):
     uinfo = db(db.auth_user.username == username).count()
     if uinfo > 0:
         errors.append(f"Username {username} already exists on line {line}")
-        row = db(db.auth_user.username == username).select(db.auth_user.email,
+        user_row = db(db.auth_user.username == username).select(db.auth_user.email,
         db.auth_user.first_name, db.auth_user.last_name)
-        if (row.auth_user.email == email) and 
-        (row.auth_user.first_name == fname) and (row.auth_user.last_name == lname):
-        errors.append(f"User {username} already exists on line {line}")
+        if (user_row.auth_user.email == email) and 
+        (user_row.auth_user.first_name == fname) and (user_row.auth_user.last_name == lname):
+            errors.append(f"User {username} already exists on line {line}")
         
     if password == "":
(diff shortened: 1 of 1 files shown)


bnmnetp (changes requested):
Hi,

The username is the unique identifier for the student throughout runestone.  So any check to see if the user exists must check the username.

Eaw7:
Hello, I am unfamiliar with Python's database function calls. Is there any guidance you can give me on how to correct this? This is in reference to admin.py around line 2486 in the _validateUser function.

  ```
  if uinfo > 0:
        errors.append(f"Username {username} already exists on line {line}")
        user_row = db(db.auth_user.username == username).select(db.auth_user.email,
        db.auth_user.first_name, db.auth_user.last_name)
        if (user_row.auth_user.email == email) and 
        (user_row.auth_user.first_name == fname) and (user_row.auth_user.last_name == lname):
            errors.append(f"User {username} already exists on line {line}")
```

Thank you for your time!

File: controllers/admin.py [hunk 1]
```diff
@@ -2485,7 +2485,12 @@ def _validateUser(username, password, fname, lname, email, course_name, line):
     uinfo = db(db.auth_user.username == username).count()
     if uinfo > 0:
         errors.append(f"Username {username} already exists on line {line}")
-
+        user_row = db(db.auth_user.username == username).select(db.auth_user.email,
+        db.auth_user.first_name, db.auth_user.last_name)
+        if (user_row.auth_user.email == email) and
```
bnmnetp, line 2490: This is a syntax error.  You cannot just arbitrarily split a python statement into multiple rows. You need to use a \ at the end of 2490 to indicate that you are splitting this single boolean expression onto multiple lines.

Also we use black to format our files and black would have flagged this, anything would have flagged this syntax error if you would have tried to test it.

bnmnetp: The intent of this issue is to distinguish between two types of duplicates:

1.  An instructor is re-uploading a CSV file for a class but has added more students to the class. In this case we just want to make this a warning.  This would mean that the user has the same username, email, first_name in the database. THIS should be a warning.

2. An instructor is trying to register a username for a student in their class but a different human already has that username.  THIS should be an error.

PR #1651, opened 2021-04-22
Fix: Avoid exception.

bjones1:
I found a case with a timed selectquestion tests which produces an exception here , since I don't specify the points for each question. This is the obvious fix.

### DIFF
--- a/controllers/ajax.py
+++ b/controllers/ajax.py
@@ -1670,7 +1670,8 @@ def get_question_source():
         logger.debug(
             f"Assignment Points for {assignment_name}, {selector_id} = {ui_points}"
         )
-        points = ui_points.points
+        if points:
+            points = ui_points.points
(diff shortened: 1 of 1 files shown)


bjones1:
After this is merged, then I can provide a passing PR for timed selectquestion tests!

controllers/ajax.py
```diff
@@ -1670,7 +1670,8 @@ def get_question_source():
         logger.debug(
             f"Assignment Points for {assignment_name}, {selector_id} = {ui_points}"
         )
-        points = ui_points.points
+        if points:
```
bnmnetp, line 1673: I think this should be `if ui_points:`

bjones1:
Good point. I"m closing this and merging this small fix with the commit adding tests on the server.

PR #1655, opened 2021-05-18
Changes to use the new split/bundled runestone components

bnmnetp:
@bjones1 This looks like it works with the instructor interface.

### DIFF
--- a/views/_sphinx_static_files.html
+++ b/views/_sphinx_static_files.html
@@ -13,7 +13,20 @@
 <script type="text/javascript" src="https://cdn.jsdelivr.net/npm/vega@4.0.0-rc.2/build/vega.js"></script>
 <script type="text/javascript" src="https://cdn.jsdelivr.net/npm/vega-lite@2.5.0/build/vega-lite.js"></script>
 <script type="text/javascript" src="https://cdn.jsdelivr.net/npm/vega-embed@3.14.0/build/vega-embed.js"></script>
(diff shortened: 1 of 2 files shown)


views/_sphinx_static_files.html
```diff
@@ -13,7 +13,20 @@
...
+{{ import json }}
+{{ with open("".join(["applications/",
+                       request.application,
+                       "/books/",
+                       course['base_course'],
+                       "/published/",
+                       course['base_course'],
+                       "/_static/webpack_static_imports.json"]),
+             encoding="utf-8") as f:
+      for js_import in json.loads(f.read()):
```
bjones1, line 25: This will need to change slightly with my new PR, since the JSON file now has both CSS and JS to include; otherwise, it looks good.

bnmnetp: I'm seeing a change in format of this webpack_static_imports.json file but I'm not getting any css files in it yet..??

bnmnetp:
I'm seeing lots of test failures, that don't seem related to this PR at all...  Hard to figure out with limited bandwidth, do you have any ideas?

views/_sphinx_static_files.html
bjones1: In dev mode, it generates no CSS. In production mode, it does -- this is based on webpack's recommended production mode approach (see https://webpack.js.org/guides/production/#minimize-css).

bjones1:
I'll take a look...

PR #1657, opened 2021-05-21
Runestone bundle server

bjones1:
I couldn't figure out how to revise/push to #1655, so I'm opening this instead. This ended up being a smorgasboard of fixes to work around several existing problems:

- The sphinx template html code was being included twice, because the syntax of some templates was wrong. I fixed the syntax plus indented the code to make the flow of the templates clearer.
- The selectquestion test fails, because mathjax isn't included by Sphinx on that page, since it doesn't have math. As a workaround, I added an equation to the page. However, this is a fundamental problem with selectquestion and Sphinx.

### DIFF
--- a/.github/workflows/python-app.yml
+++ b/.github/workflows/python-app.yml
@@ -51,7 +51,7 @@ jobs:
                   mkdir runestone
                   shopt -s extglob
                   mv !(runestone) runestone
-                  git clone https://github.com/RunestoneInteractive/RunestoneComponents.git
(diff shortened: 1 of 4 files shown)


bjones1:
Tests are failing, but it looks like the Postgres config is broken for some reason -- another PR from master with a trivial change to some HTML is failing.

bjones1:
I've updated my Runestone Components branch to fix these database failures.

bjones1:
This is now rebased on top of a working master branch.

PR #1670, opened 2021-06-11
For parsons distractor helptext

vqum:
In conjunction with https://github.com/RunestoneInteractive/RunestoneComponents/pull/1206


views/assignments/doAssignment.html
```diff
@@ -158,6 +158,10 @@ <h5 style='text-align:center;'>Not yet graded</h5>
     $("#gradeMeButton").on("click", function(){
         $('#gradeMeButton').css('visibility', 'hidden');
         selfGrade({{=assignment['id']}}, '{{=student_id}}')});
+
```
bnmnetp, line 161: What is this for?  It seems like it would just toggle all tooltips for everything on the assignment page??

Issue #1715, opened 2021-07-26
Reorganize Runestone Library 


qasema:
The Runestone Library should be reorganized in a way that books are accessed easily and shows more descriptions of each book. It should also include search bar. That will make easier for the user to find the books they want by typing the title of the book or some keywords.   

This is the initial design:
 
![Capture](https://user-images.githubusercontent.com/68142980/127046768-0046de0d-f87b-4645-8f86-b7b64ebac774.PNG)

PR #1716, opened 2021-07-26
Library

qasema:
fix issue #1715: 
 
Reorganized Runestone Library, add titles and descriptions to the books, add search bar. 

@bnmnetp each conf.py should include two variable named : _course_description_   and _key_words_ . for example:

course_description = " description ......."
key_words = " fopp21 Brad Miller fall_21 name_college ....."


models/grouped_assignments.py
```diff
@@ -7,7 +7,7 @@
     Field("course", db.courses),
     Field("name", "string"),
     Field(
-        "points", "integer", default=0, notnull=True
```
bnmnetp, line 10: This needs to be put back.  I can walk you through updating your database so it won't cause you problems.

bnmnetp (changes requested):
If black is still causing you trouble it may be that you need the version that matches our unit tests.  That is version 20.8b1. You can run black --version to see which one you have.

qasema:
sorry! the last two commits are: fix  black_format_check &  fix group_assignments.py (some typo)

bnmnetp (approved):

Issue #1717, opened 2021-07-26
New Front Page


Bryar-Frank:
As part of a larger project to change the flow of initial sign up that Brad has suggested to a team, it has come to attention that the user page hasn't been updated in a long time and needs to be redone.

The start of this process is to create a new front page that is separate from the user.html page.
The login and registration will look a bit different and this implementation should be able to fit into the current plane without messing up any of the active logic.

Bryar-Frank:
The pull request just submitted is mostly complete and commented, but needs a little more reworking for the button layout after person is already logged in.

Bryar-Frank:
I have now reworked the buttons. The buttons for the user after they are logged in will have to be reworked once everything else with the registration is complete, but since users aren't assigned as instructors/students/or other yet, I cannot define the layout of the buttons per user until then. Otherwise, this implementation works completely with the current version of Runestone.

Bryar-Frank:
I have closed the Front Page pull request since I merged the "Front Page" branch with "Other Register". I will be linking the new branch is just a bit that will also add in the registration route for anyone signing up that is not a student or instructor.

PR #1718, opened 2021-07-26
Front page

Bryar-Frank:
starts fix for issue #1717 

This is mostly complete, but still needs to be reworked a little bit with the options for when a person is already logged in.


Bryar-Frank:
I have now reworked the buttons. The buttons for the user after they are logged in will have to be reworked once everything else with the registration is complete, but since users aren't assigned as instructors/students/or other yet, I cannot define the layout of the buttons per user until then. Otherwise, this implementation works completely with the current version of Runestone.

Bryar-Frank:
@bnmnetp

Bryar-Frank:
This pull request is being closed and new one will be opened

Issue #1730, opened 2021-08-05
JS error in dynamically loading activecode


bjones1:
In [views/layout.html line 282](https://github.com/RunestoneInteractive/RunestoneServer/blob/master/views/layout.html#L282) there's the old method to load activecode. I remember @bnmnetp fixed this nicely for the Runestone Components -- would you apply the same fix here?

bjones1:
Aack, this is more complicated. It looks like we don't load the RS components JS in `layout.html`. Could we do that? It feels somewhat like a re-thinking of e41bc40f, where we put code from `_sphinx_static_files.htm` in `layout.html`. @bnmnetp, what do you think?

bjones1:
...or do we omit the pencil / activecode popup icon on login pages? If we leave this in book-only pages, it's an easy fix...

bnmnetp:
I think scratch activecodes work on all of the instructor dashboard pages, just not the login pages.  I think we could safely omit it from the login pages, but I would keep it available on other pages.  Which I think probably translates to not showing it on any page that uses the default/user controller.

PR #1733, opened 2021-08-06
Fix: Don't show scratch activecode in the user view, since RS components aren't loaded there.

bjones1:
Per the discussion in #1730.


psmaker:
Just a clarification what "user view" means - so is this saying that the Scratch ActiveCode will not work in the Assignment page or any assignments that are made? I notice this behavior and thought it was a bug that the Scratch ActiveCode does not work on certain pages. (I don't know why I am just noticing this behavior this year though...)

If that is the case, it would be nice if the little Scratch ActiveCode pencil icon does not show up on those pages since it by design does not do anything on those pages.

PR #1736, opened 2021-08-09
Fix: Switch to using the myst parser, since recommonmark is deprecated.

bjones1:


bnmnetp:
I think this needs to be updated so that it includes the previous changes.  Why GitHub isn't giving me the button to bring this up to date with master is a mystery.

bjones1:
Thanks for helping track this! I merged in master, and hopefully that will do it.

bnmnetp (approved):

PR #1739, opened 2021-08-10
Bookserver

bnmnetp:
Updates to the web2py server to make it work with the new bookserver.


bnmnetp:
@bjones1 here is a branch for changes to the web2py server to make it work alongside the new bookserver.  This first push includes my changes to create an access token.

controllers/default.py
```diff
@@ -175,6 +177,16 @@ def index():
             % (request.application, request.application)
         )
     else:
+        # At this point the user has logged in
+        # add a jwt cookie for compatibility with bookserver
+        token = _create_access_token(
+            {"sub": auth.user.username}, expires=datetime.timedelta(hours=24 * 30)
```
bjones1, line 183: Nitpick: how about `timedelta(days=30)`?

controllers/default.py
```diff
@@ -175,6 +177,16 @@ def index():
...
+        )
+        # set cookie
+        response.cookies["access_token"] = token
+        response.cookies["access_token"]["expires"] = 24 * 3600 * 90
```
bjones1, line 187: Why are the expires times different here (90 days, vs 30 days a few lines ago)? Does it matter?

controllers/default.py
```diff
@@ -610,3 +622,40 @@ def delete():
...
+        expires_in = datetime.datetime.utcnow() + datetime.timedelta(minutes=15)
+
+    to_encode.update({"exp": expires_in})
+
+    if scopes is not None:
+        unique_scopes = set(scopes)
+        to_encode.update({"scopes": list(unique_scopes)})
+
+    algorithm = "HS256"  # normally set in constructor
+    secret = "supersecret"  # set in settings.secret
```
bjones1, line 658: Hmm, do we want to import this from the bookserver settings eventually? It looks like we'll need to have web2py running side-by-side web bookserver for a while.

controllers/default.py
```diff
@@ -610,3 +622,40 @@ def delete():
         auth.logout()  # logout user and redirect to home page
     else:
         redirect(URL("default", "user/profile"))
+
+
+def _create_access_token(data: dict, expires=None, scopes=None) -> str:
```
bjones1, line 627: This is awesome! Can you point to any FastAPI docs that discuss the expected token, or did you discover this by looking at the cookies FastAPI uses?

bjones1 (approved):
Good work!

controllers/default.py
bnmnetp: I looked at the code for the fastapi_login plugin

controllers/default.py
bnmnetp: We need to think about the expiration stuff.  The fastapi_login plugin respects what is in the encrypted token.  So that is one level of expiration.  You are right that it makes no sense to have the cookie expiration longer than the token expiration.    But we definitely will want to have tokens lasting more than 12 hours.

controllers/default.py
bnmnetp: I think the natural place to set that for web2py is in models/1.py

I guess it shows that when I started out on this, yesterday it was a total experiment.  but the experiment worked more quickly than I expected.  So we have a few things to iron out before we call it production ready.

controllers/default.py
bnmnetp: I will definitely add a comment to acknowledge that it is something of a ripoff of one of their methods.

controllers/default.py
bjones1: That makes sense. It's amazing how quick this is coming together!

controllers/default.py
bjones1: Sounds good!

bnmnetp:
Ugh, tests are passing locally but every one of them is generating a ticket on the test action...
We need a way to report that error when it happens

bnmnetp:
No idea what causes this but on the dev server I am now getting:

```
  File "/srv/web2py/gluon/custom_import.py", line 78, in custom_importer
    result = sys.modules[modules_prefix]
KeyError: 'applications.runestone.modules.jwt'

  File "/srv/web2py/gluon/restricted.py", line 219, in restricted
    exec(ccode, environment)
  File "/srv/web2py/applications/runestone/controllers/default.py", line 13, in <module>
    import jwt  File "/srv/web2py/gluon/custom_import.py", line 80, in custom_importer
    raise ImportError("No module named %s" % modules_prefix)
ImportError: No module named applications.runestone.modules.jwt
```
even though pyjwt is clearly installed and I can import it from the cl

bnmnetp:
Ugh, I hate web2py's custom_import functionality.

Here is what was happening:  jwt tries to import cryptogography inside a try block.  It catches the exception and sets has_crypt to False.   But the custom importer throws a different uncaught exception. so jwt's attempt to catch the exception is foiled.  This ends up showing in the ticket as the above error message!

The solution was to add cryptography as a module to requirements.in

Not sure how or why but on my local setup crtypography was installed.

bjones1:
Wow, I will really miss working with web2py... :)

PR #1746, opened 2021-08-25
Db to bookserver

bjones1:
Fixes (plus one DELME) to get web2py playing nicely with the bookserver.

### DIFF
--- a/models/0.py
+++ b/models/0.py
@@ -33,7 +33,6 @@ settings.server_type = "http://"
 settings.academy_mode = True
 settings.lti_only_mode = False
 settings.coursera_mode = False
-settings.secret = "supersecret"
 
 # Do not control this with hostnames
 config = environ.get("WEB2PY_CONFIG", "NOT SET")
--- a/models/0.py
+++ b/models/0.py
@@ -33,6 +33,7 @@ settings.server_type = "http://"
 settings.academy_mode = True
 settings.lti_only_mode = False
 settings.coursera_mode = False
+settings.secret = "supersecret"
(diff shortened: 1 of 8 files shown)


bjones1:
@bnmnetp, I'm confused -- this got marked as closed because of a merge in another repo? Is this github weirdness? This PR still definitely needs to be merged...

bnmnetp:
It is still in my open PR list, it just needs to have a conflict resolved.

Separate but related:

I wonder if the flow that we use in entrypoint.sh for the docker container where we use the rsmanage command to check the state of the database and if necessary create all of the tables before starting EITHER server  might not give us a little more flexibility in the short/medium term than completely disabling migrations?  I only say this because I think for both of us (and others running servers) the main use case is going to be to keep the existing database and have bookserver work with it.

bjones1:
Oops, I missed the conflict. I went ahead and merged in upstream/bookserver and pushed the change.

I agree that we need to plan for both servers running together in the short/medium term. I'll discuss with you on Slack.

bjones1:
@bnmnetp , this is ready now. Let me know if you see any other changes you want me to make.

bnmnetp:
Thanks @bjones1 -- Is this holding up anything for you?  I have those two changes to the schema that are part of my peer_support branch.  So I'm thinking whether I should wait on this until after that is merged or if I should just go ahead and merge this and deal with that migration when I merge that branch.

bjones1:
This no longer has schema changes, just disables migrations for bookserver tables.

PR #1753, opened 2021-09-09
Fixes issue 1728. Adds spacing margin in index.html

### LINKED ISSUE #1728: Runestune Library spacing
The Runestone library needs some spacing an alignment https://runestone.academy/runestone/books/index

zachneill:
Fixes #1728 

Added spacing to views/books/index.html 
To do this, added style="margin: 25px" which adds a margin in between books.

### DIFF
--- a/views/books/index.html
+++ b/views/books/index.html
@@ -20,7 +20,7 @@
         
             {{ for book in book_list: }}
             
-            <div id='list' class='library_entry'>
+            <div style="margin: 25px;" id='list' class='library_entry'>
                 <div class="book_title"> {{=book['title']}}
                     <a href="{{=book['url']}}"> 
                     <span class="link1"></span> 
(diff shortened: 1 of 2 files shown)


views/books/index.html
```diff
@@ -20,7 +20,7 @@ <h2>Runestone Interactive Library of Books</h2>
         
             {{ for book in book_list: }}
             
-            <div id='list' class='library_entry'>
+            <div style="margin: 25px;" id='list' class='library_entry'>
```
bnmnetp, line 23: Hi,

This would be better if it was in the `runestone-custom-sphinx-bootstrap.css` file.  I know we don't always set a great example, but we are working on getting styling out of html and into css as much as we can.

bnmnetp (changes requested):

zachneill:
Is this a good way to go? I added a 25px margin in the library_entry class in index.css. Instead of using styles.

import-brain (approved):
LGTM!

@bnmnetp, how is it now?

PR #1761, opened 2021-09-20
Fix to payments

bjones1:
Fix: If a student already owns a course, then any courses from the same base course are free.


bjones1:
Closing until I get CI to pass.

PR #1774, opened 2021-09-28
Docker updates

bjones1:
This is a rewrite of the Docker approach, so that we can run the Runestone server along with the BookServer, plus provide more flexible container creation options. **A neat feature:** if we distribute this script plus the `ci_tools.py` script it needs, then it can download the Runestone repo, installer docker, then build the container -- a very simple process.

The main change is to do all Docker build/run from a single Python script, which makes it easier to manage this complexity. Instead of `docker build`, use `docker/docker_tools.py`:

```
Usage: docker_tools.py build [OPTIONS]

  When executed outside a Docker build, build a Docker container for the
  Runestone webservers.

  Inside a Docker build, install all dependencies as root.

Options:
  --arm / --no-arm       Install the ARMv7 toolchain.
  --dev / --no-dev       Install tools needed for development with the
                         Runestone.

  --pic24 / --no-pic-24  Install tools needed for development with the
                         PIC24/dsPIC33 family of microcontrollers.

  --rust / --no-rust     Install the Rust toolchain.
  --tex / --no-tex       Instal LaTeX and related tools.
  --help                 Show this message and exit.
```

I'd appreciate any feedback and testing! I know that @ascholerChemeketa has worked on Docker stuff recently, and that @bnmnetp has a proto of this already running, so your help would be much appreciated. There's still lots of testing needed, along with suggestions for improving the overall process.


bjones1:
This is ready -- it's working well for me.

One thing to think about: DBURL naming. We have a dev, production, and test DBURL for sync (web2py) and async (FastAPI). It's painful to keep all this in sync. How about setting FastAPI to use the sync flavor of DBURL then internally making it async?

docker/docker_tools.py
```diff
@@ -0,0 +1,541 @@
...
+def build(arm: bool, dev: bool, passthrough: Tuple, pic24: bool, tex: bool, rust: bool):
+    """
+    When executed outside a Docker build, build a Docker container for the Runestone webservers.
+
+        PASSTHROUGH: These arguments are passed directly to the underlying "docker build" command. To pass options to this command, prefix this argument with "--". For example, use "docker_tools.py build -- -no-cache" instead of "docker_tools.py build -no-cache" (which produces an error).
+
+    Inside a Docker build, install all dependencies as root.
+    """
+
+    assert is_linux, "This program is only capable of running on Ubuntu/Debian."
```
bnmnetp, line 90: The only impediment to running on macos is removing this line. 😄

bjones1: Wow, fantastic! Fixed with pleasure.

bnmnetp:
I just merged this with the peer instruction branch.

PR #1778, opened 2021-10-01
update.gitignore

swami-raj:


bnmnetp:
removing the comment makes it less clear why we have this.

PR #1779, opened 2021-10-04
Fix: Docker improvements.

bjones1:
Per the Slack discussion.

This also includes the docs updates in ed0f08c2 that somehow got lost in the merge.


bjones1:
I've tested this from a clean Ubuntu install: download only `docker_tools.py` and run it. It will install all prereqs and build a docker image. It's a perfect easy start for students/developers wanting to get involved!

Issue #1798, opened 2021-11-10
certbot issues


bnmnetp:
Some things we need to clean up about the certbot implementation for the new docker.

1. I don't think many of the changes to the `sites-available/runestone` file from lines 637-676 are really necessary.  We should set up the server_name correctly.  BUT, certbot will do the rest of the work for us to redirect http to https as well as modifying the file with the appropriate certificate information.
2. The certbot command that that gets the certificate and saves them into `/etc/letsencrypt` should only be done during the build -- NOT on every startup.
3. Here is the version of the file that works for me:  -- note the certbot additions are commented as `# managed by certbot`

```
 *******************************                                                                                          
# |docname| - nginx configuration                                                                                          
# *******************************                                                                                          
# This sets up nginx to run both the old and new server together. It requires `processing by Python <Set up nginx based on 
env vars.>`_ before use.                                                                                                   
#                                                                                                                          
# This file was partially taken from the `gunicorn deployment docs <https://docs.gunicorn.org/en/stable/deploy.html#nginx-c
onfiguration>`_.                                                                                                           
#                                                                                                                          
# `gzip_static <http://nginx.org/en/docs/http/ngx_http_gzip_static_module.html>`_: Enables ("on") or disables ("off") check
ing the existence of precompressed files. Send compressed files in lieu of uncompressed files where they exist. This requir
es the optional ``ngx_http_gzip_static_module`` module.                                                                    
gzip_static on;                

# Define the web2py and FastAPI servers.                                                                                   
server {                                                                                                                   
    # For testing, use port 80 (http); for production, use port 443 (https).                                               
                                                                                                                           
    include /etc/nginx/default.d/*.conf;                                                                                   
                                                                                                                           
    # `server_name <http://nginx.org/en/docs/http/ngx_http_core_module.html#server_name>`_: Set name(s) of a virtual server
.                                                                                                                          
    server_name dev.runestoneacademy.org;                                                                                  
                                                                                                                           
    # Rewrite the path to static files.                                                                                    
    #                                                                                                                      
    # `location <https://nginx.org/en/docs/http/ngx_http_core_module.html#location>`_: set configuration depending on a req
uest URI. The ``~*`` indicates the following parameter is a case-insensitive regex. Look for web2py static files.          
    #                                                                                                                      
    # Match web2py static paths with the application specified, such as ``/runestone/static/``.                            
    #                                                                                                                      
    # root supplies the uptree path the match gets appended to the path specified                                          
[...]

PR #1799, opened 2021-11-10
Docker

bjones1:
This adds an authoring option and fixes certbox problems per #1798.

### DIFF
--- a/docker/docker_tools.py
+++ b/docker/docker_tools.py
@@ -789,6 +789,9 @@ def _build_phase2(
             xqt(
                 'certbot -n --agree-tos --email "$CERTBOT_EMAIL" --nginx --redirect -d "$RUNESTONE_HOST"'
             )
+        else:
+            # Renew the certificate in case it's near its expiration date. Per the `certbot docs <https://certbot.eff.org/docs/using.html#renewing-certificates>`_, ``renew`` supports automated use, renewing only when a certificate is near expiration.
+            xqt("certbot renew")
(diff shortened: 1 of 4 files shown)


docker/docker_tools.py
```diff
@@ -800,9 +784,11 @@ def _build_phase2(arm: bool, dev: bool, pic24: bool, tex: bool, rust: bool):
 
     # Certbot requires nginx to be running to succeed, hence its placement here.
     if env.CERTBOT_EMAIL:
-        xqt(
-            'certbot -n --agree-tos --email "$CERTBOT_EMAIL" --nginx --redirect -d "$RUNESTONE_HOST"'
-        )
+        # See if there's already a certificate for this host. If not, get one.
+        if not Path(f"/etc/letsencrypt/live/{env.RUNESTONE_HOST}").is_dir():
```
bnmnetp, line 788: We **may** want to think about running `certbot renew` if the cert exists to avoid any staleness due to this being in a volume.

bnmnetp (approved):

docker/docker_tools.py
bjones1: Per the [docs](https://certbot.eff.org/docs/using.html#renewing-certificates), "renew is suitable (and designed) for automated use, to allow your system to automatically renew each certificate when appropriate. Since renew only renews certificates that are near expiry it can be run as frequently as you want - since it will usually take no action."

I'll go ahead and add this -- good idea.

PR #1802, opened 2021-11-15
Updated Poetry

bjones1:
A number of fixes and improvements to the Poetry setup -- extends #1801.


bnmnetp (approved):
Tons of changes here that were also in #1801 -- force pushing??

This always makes me nervous.

PR #1803, opened 2021-11-15
Updated Docker

bjones1:
This PR builds on #1802 by changing Docker to use the new Poetry config.

It also uses gunicorn instead of uwsgi for web2py.


bnmnetp (commented):
So much duplicated from the previous two PRs that its hard to know, exactly what's new.

There are a LOT of good maintenance things in the `runestone.ini` file that are worth thinking about before they get lost.

1. touch file to reload everything
2. automatic restarting of worker processes after X number of requests or X amount of memory usage - to keep the python processes from growing to unbounded size.
3. automatically killing stuck processes

All of these were painstakingly discovered after having the server blow up over and over.  I would like to not lose that as I don't want to go through that under enormous pressure at the start of spring semester.

docker/nginx/sites-available/runestone.template
```diff
@@ -61,13 +61,17 @@ server {
         # `proxy_pass <https://nginx.org/en/docs/http/ngx_http_proxy_module.html#proxy_pass>`_: set the protocol and address of a proxied server and an optional URI to which a location should be mapped. **Tricky**: Specifying the ``location`` above with a trailing slash and including a trailing slash at the end of ``proxy_pass`` causes nginx to strip of the ``/ns`` prefix when sending it to gunicorn. Quoting the docs: "If the ``proxy_pass`` directive is specified with a URI, then when a request is passed to the server, the part of a normalized request URI matching the location is replaced by a URI specified in the directive."
         #
         # _`gunicorn socket`: This matches the ``--bind`` parameter when `gunicorn is run <run_bookserver>`.
-        proxy_pass http://unix:/run/gunicorn.sock:/;
+        proxy_pass http://unix:/run/fastapi.sock:/;
     }
 
     location / {
-        # Per `uwsgi_param <https://nginx.org/en/docs/http/ngx_http_uwsgi_module.html#uwsgi_param>`_, this passes standard CGI environment variables.
```
bnmnetp, line 68: I assume this is for web2py?  This is where the /runestone/ part of the URL actually helps and makes things clearer.

models/db.py
```diff
@@ -213,7 +213,7 @@ def get_course_url(*args):
 
 
 def getCourseNameFromId(courseid):
-    """ used to compute auth.user.course_name field """
```
bnmnetp, line 216: New version of black?

Dockerfile
```diff
@@ -32,12 +32,10 @@ ARG DEBIAN_FRONTEND=noninteractive
 # We need the entire Runstone server repo for the build.
 COPY . ${RUNESTONE_PATH}
 RUN \
```
bnmnetp, line 34: OK, so if we have DOCKER_BUILD_ARGS set -- we can still use `docker build....` right?

docker/gunicorn/web2py.conf.py
```diff
@@ -0,0 +1,44 @@
...
+# `user <https://docs.gunicorn.org/en/stable/settings.html#user>`_: Switch worker processes to run as this user.
+user = "www-data"
+
+# `group <https://docs.gunicorn.org/en/stable/settings.html#group>`_: Switch worker process to run as this group.
+group = "www-data"
+
+# `workers <https://docs.gunicorn.org/en/stable/settings.html#workers>`_: The number of worker processes for handling requests. Pick this based on CPU count.
+workers = multiprocessing.cpu_count()
+
+# `bind <https://docs.gunicorn.org/en/stable/settings.html#bind>`_: The socket to bind. This must match the socket nginx sends to.
```
bnmnetp, line 43: We should add 

`max_requests = 500` or more not sure but in uwsgi I have it at 500 - this is a good way to prevent memory "leaks" in Python to keep things from eventually crashing.
`max_requests_jitter = 30` I pulled 30 out of my hat. but it ensures that everything doesn't restart in lock step.
`timeout=60` - like the harakiri option for uwsgi this restarts processes that haven't produced any output for 60 seconds.

bnmnetp (changes requested):

bnmnetp:
In addition to the comments in the review it appears that #1802 is broken.  It tries to install from `RunestoneServer/requirements-dev.txt` which no longer exists.

Dockerfile
bjones1: It's possible to invoke `docker-build` directly, but does require an extra param (the `DOCKER_BUILD_ARGS`).

docker/gunicorn/web2py.conf.py
```diff
@@ -0,0 +1,44 @@
...
+wsgi_app = "wsgihandler:application"
+
+# `user <https://docs.gunicorn.org/en/stable/settings.html#user>`_: Switch worker processes to run as this user.
+user = "www-data"
+
+# `group <https://docs.gunicorn.org/en/stable/settings.html#group>`_: Switch worker process to run as this group.
+group = "www-data"
+
+# `workers <https://docs.gunicorn.org/en/stable/settings.html#workers>`_: The number of worker processes for handling requests. Pick this based on CPU count.
+workers = multiprocessing.cpu_count()
```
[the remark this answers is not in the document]
bjones1: I agree -- please change this based on your experience. (Or tell me what to use and I'll change it.)

docker/gunicorn/web2py.conf.py
bjones1: Fantastic, thanks. Will do!

bjones1: I'm putting these in a shared config between web2py and fastapi.

models/db.py
bjones1: Yes, Python 3.9.8 breaks the older version of Black and causes CI to fail.

bjones1:
> So much duplicated from the previous two PRs that its hard to know, exactly what's new.

I agree, this is painful. I was hoping that merging PRs would make this clearer. I could rebase it that would help.

> There are a LOT of good maintenance things in the `runestone.ini` file that are worth thinking about before they get lost.
> 
> 1. touch file to reload everything
> 2. automatic restarting of worker processes after X number of requests or X amount of memory usage - to keep the python processes from growing to unbounded size.
> 3. automatically killing stuck processes
> 
> All of these were painstakingly discovered after having the server blow up over and over. I would like to not lose that as I don't want to go through that under enormous pressure at the start of spring semester.

I agree. I'm hope to any improvements. I assume you prefer to use gunicorn for both FastAPI/BookServer and web2py, instead of gunicorn for FastAPI/BookServer (no option on this one) and uwsgi for web2py? This seems like simpler operation long-term (tuning one server instead of two).

bjones1:
> In addition to the comments in the review it appears that #1802 is broken. It tries to install from `RunestoneServer/requirements-dev.txt` which no longer exists. See lines 549-557 in docker_tools.py.
 
I can't find this in my source - ???

> I think we need to update/or add a workflow on GitHub that tests the docker configuration.

I agree. I'll look at this.

Dockerfile
bjones1: ...but the build process requires the entire repo to be present, so this is probably a non-starter...

docker/gunicorn/web2py.conf.py
bjones1: Are you sure on 500 max requests? That seems like a very small amount of work before a restart...

bnmnetp:
If you look at docker_tools.py on master here on GitHub you will see it.  So maybe you have something that is not pushed?

docker/nginx/sites-available/runestone.template
bjones1: Yes, only `/ns` routes to the BookServer; everything else goes to web2py.

bjones1:
Thanks, I see that looking on Github master. It's fixed in this PR.

Issue #1775, opened 2021-09-30
Automating Development Enviorment Setup


Cortesj:
Runestone appears to be quite difficult to develop for in-terms of getting the development environment up and running. I believe there can be something done to update scripts such as “dbuild” and the main docker file to better accommodate the building of books and avoid errors whenever possible, within the build process. I have some concepts/ideas of adding functionality or scripts to assist in the build/setup process. I am interested on contributing some of my ideas towards achieving this goal.

Note: I am Berea College student and will be currently evaluating this issue and how to specifically implement/fix said issue.

bnmnetp:
Great!

the first thing you could try that would be very useful is to check out PR #1774 and give it a try!  We will want to build off of that work.

Brad

bjones1:
@cortesj, when you're ready, I can provide some suggestions on how to improve the process. For a quick start, run `curl -fsSLO https://raw.githubusercontent.com/bjones1/RunestoneServer/docker_updates/tests/docker_tools.py | python3 -` from any Linux distro with `curl` installed.

Cortesj:
@bjones1 I've actually been looking through your python script and it looks great so far! What suggestions did you have in mind?

bjones1:
@Cortesj, thanks for the feedback. I'd suggest some integration work: we have a bunch of scripts `scripts/dshell`, etc. These could all be made part of the `docker_tools.py` script -- `docker_tools.py shell`, for example. Also, do they all still work after changing everything? Some testing would be helpful. I can be more specific on how to accomplish these goals if you'd like.

PR #1804, opened 2021-11-16
Adds online functionality to the book-build cmd & requirement's check.

Cortesj:
Adds additional functionality to docker_tools. Two additional parameters added "--online" & "--username" which means that specified book will download from online repository if "--online" flag is specified. The "--username" pulls from specified GitHub Username. The appropriate checks have been added to check for existing repositories/books and verify that remote repository is a valid repository. If username specified doesn't have valid repository then default to _RunestoneInteractive_ as default username to attempt cloning the book.

Additional functionally added is checking if book has any requirements prior to build & deployment. It will attempt to find a requirements.txt file within sub_book_name. If nothing is found then it will skip step and proceed with build & deployment.

**TL:DR**
- Adds requirements.txt check to book-build
- Adds flag "--online" to clone book from Remote Repository
- Adds flag "--username" to clone book from specific GitHub user when "--online" is specified
- Check if remote repository is available and following certain actions in case of error reaching remote repo.

Pertains to issue #1775

### DIFF
--- a/docker/docker_tools_misc.py
+++ b/docker/docker_tools_misc.py
@@ -80,12 +80,36 @@ def run_bookserver(dev: bool) -> None:
 # --------------
 @click.command()
 @click.argument("book_sub_name")
-def book_build(book_sub_name) -> None:
+@click.option(
+    "--online/--no-online", 
+    default=False,
+    help="Download repo from an online Github Repository",
+)
+@click.option(
+    "--username",
+    default="RunestoneInteractive",
+    nargs=1,
+    help="Clone repository with specified Github Username",
+)
+def book_build(book_sub_name, online: bool, username: str) -> None:
     """
     Build a Runestone e-book, where BOOK_SUB_NAME provides the name of the subdirectory where a book resides.
+    
+    Optionally: Use the 'Online' and 'Username' flags to specify if you wish to download a repository and subsequently build it.
     """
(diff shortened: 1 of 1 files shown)


bjones1:
Jose,

I'd suggest adding this functionality to `rsmanage` instead of the Docker tools -- that way, users can install books on a stand-along version (BookServer only) or on the combined web2py + BookServer setup.

bnmnetp:
@bjones1 - I don't understand your comment, building a book is part of the docker_tools.py script.  It is not part of rsmanage.

I know you have proposed, elsewhere, that we migrate the build command to rsmanage but that hasn't happened.

Cortesj:
@bjones1 Could you explain more on rsmanage? I haven't seen that script yet. I am assuming its currently within the master build. I will be taking a look at it to see how I can implement it there.

bjones1:
@Cortesj, sorry, I missed this earlier. Take a look at #1810 -- does that help? Also, have you joined https://runestoneacademy.discourse.group/? There's a fair amount of discussion there.

Cortesj:
I noticed that a discourse group was created and joined it fairly recently. I guess I have more reading to catch up on! I might close the PR considering that movement to different functionality to different portions of the system.

bjones1:
I agree -- I really like this code, but it belongs in `rsmanage`.

PR #1806, opened 2021-11-16
Pre bookserver recaptcha

schweinsbergs:
Implements reCaptcha v2. The user needs to complete it before they continue in the sign up process. 

You'll need to implement your own keys. For right now, I generated my own reCaptcha keys to test.  **The current key will only work on Luis and I's droplet and runestone.academy**. Line 298 in user.html can be swapped out with your own keys and you can add what domains will be valid.

### DIFF
--- a/views/default/user.html
+++ b/views/default/user.html
@@ -221,10 +221,14 @@
                         <script type="text/javascript">
                             function checkTcp() {
                                 let cb = document.getElementById("auth_user_accept_tcp");
-                                if (cb.checked) {
+
+                                var reCaptchaKey = grecaptcha.getResponse();
+                                let reCaptchaValidation = Boolean(grecaptcha.getResponse());
+
+                                if (cb.checked && reCaptchaValidation) {
                                     return true;
                                 } else {
-                                    alert("You must accept our privacy policy to register.");
+                                    alert("You must accept our privacy policy and complete reCaptcha to register.");
                                     return false;
                                 }
                             }
@@ -294,8 +298,6 @@
                                                     <div class="g-recaptcha" data-sitekey="6LdiuxodAAAAAL8EqAlBnVsY91SJyFK-hKCITDM8"></div>
                                                     <br/>
                                                   </form>
-                                              
-
                                             </div>
                                         </td>
                                         <td class="w2p_fc"></td>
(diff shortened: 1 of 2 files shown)


views/default/user.html
```diff
@@ -221,10 +221,14 @@ <h2>Runestone Registration</h2>
                         <script type="text/javascript">
                             function checkTcp() {
                                 let cb = document.getElementById("auth_user_accept_tcp");
-                                if (cb.checked) {
+
+                                var reCaptchaKey = grecaptcha.getResponse();
```
bnmnetp, line 225: This does not look right to me.  You care calling the `getReponse` method two times in two lines of code.  If that is really how it is supposed to work please ad a comment to explain why this is correct.

This also adds the reCaptcha functionality unconditionally?  I would prefer that it remains an option.

bnmnetp (changes requested):
See my comments around line 225

views/default/user.html
schweinsbergs: We're not quite sure what you mean in your first comment. The way that it works currently is that it checks to see if there is a token issued by reCaptcha, and if there is, then it passes and the user is allowed to go on. Basically reCaptcha is a true or false sort of deal, the user passes or fails. 

As for the unconditional part, the option to toggle reCaptcha on or off is still in 1.py. If it's set to true, reCaptcha will be enabled on the page. If it's false, it won't be. I checked to make sure in our build environment if having it disabled and going through would cause any problems, but it seems to work as intended.

bnmnetp: The way it is written is super confusing, unless you are very familiar with how the captcha code works.

lines 225 and 226:

```
var reCaptchaKey = grecaptcha.getResponse();
let reCaptchaValidation = Boolean(grecaptcha.getResponse());
```

Line 225 looks like you are getting a key and the next line looks like you are calling get response to get something that is a boolean.  To my eye this looks like a mistake.  If the intention is to simply see if you received a key then you should document that or just use `Boolean(reCaptchaKey)` or `reCaptchaKey !=== ""`

schweinsbergs:
I see what you mean by that now, thank you for clarifying! I made that Boolean change that you mentioned.

PR #1808, opened 2021-11-17
Additional Docker commands

bjones1:
This should now provide all the commands that we need in `docker-tools`.


docker/docker_tools_misc.py
```diff
@@ -23,88 +23,72 @@
...
 # Globals
 # =======
-# The name of the container running the Runestone servers.
-res = subprocess.run(
-    'docker ps --filter "ancestor=runestone/server"  --format "{{.Names}}"',
-    shell=True,
-    capture_output=True,
-    text=True,
-)
-RUNESTONE_CONTAINER_NAME = res.stdout.strip()
```
bnmnetp, line 37: I left this as a global thinking it would be useful for several commands that we would like to migrate to the tools, such as accessing the database etc.

bnmnetp (commented):
perhaps we should wait to remove the scripts until we have actually implemented replacements.

docker/docker_tools_misc.py
bjones1: Since this means we execute `docker` for EVERY command, I tried to avoid this where possible.

The "run `psql`" command should, IMHO, be in `rsmanage` since it applies to BookServer lite (no Docker) or the Docker setup. This is somewhat a design decision, so I'll post in the newsgroup -- it seems like a better place than here, where it's somewhat incidental to this PR.

bjones1:
How about keeping a list in the newsgroups of commands to port instead, with comments/notes as appropriate?

This is already partly present in [one of my posts](https://runestoneacademy.discourse.group/t/proposal-for-admin-tools/12/4).

bnmnetp:
I don't like that approach because it makes me less productive in the meantime.  And getting the production system up and tested and stable and ready for production is in my opinion a much higher priority.

bjones1:
Now that #1807 is merged, this is ready. It completes all the docker tools we need, AFAIK.

PR #1809, opened 2021-11-18
Fixes exercisemetrics links 

### LINKED ISSUE #1765: Do not create an href for problems not supported by exercisemetrics
Currently the grading summary makes all of the problems in an assignment linkable to the exercisemetrics page.  However activecode, quizly, khan, poll, short answer, questions are not supported on that page.  

Update the get_summary API call in assignments.py so that it does not create html links for those question types.

bisterj:
This PR fixes #1765 

We also added documentation to the function get_summary 

Image below shows unsupported types not being linked

image

We are Berea College students and this pull request was approved by @pearcej

### DIFF
--- a/controllers/assignments.py
+++ b/controllers/assignments.py
@@ -155,13 +155,11 @@ group by chapter, name""",
         as_dict=True,
     )
 
-    """ List of question types that are not supported by the exercisemetrics page
-       we look through the questions table for the questions with names equal to row["name"]"""
+    """ List of question types that are not supported by the exercisemetrics page """
 
     unsupported_question_types = ['activecode', 'quizly', 'khanex', 'poll', 'shortanswer']
 
     for row in res:
-        # question = db(db.questions.name == row["name"]).select().first()
         if row["question_type"] not in unsupported_question_types: 
             if row["count"] > 0:
                 row[
--- a/controllers/assignments.py
+++ b/controllers/assignments.py
@@ -147,11 +147,10 @@ def get_summary():
     )
 
     res = db.executesql(
-        """
-    select chapter, name, min(score), max(score), to_char(avg(score), '00.999') as mean, count(score) from assignment_questions join questions on question_id = questions.id join question_grades on name = div_id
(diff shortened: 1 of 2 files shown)


controllers/assignments.py
```diff
@@ -145,6 +145,7 @@ def get_summary():
         .select()
         .first()
     )
+
     res = db.executesql(
         """
     select chapter, name, min(score), max(score), to_char(avg(score), '00.999') as mean, count(score) from assignment_questions join questions on question_id = questions.id join question_grades on name = div_id
```
bnmnetp, line 151: If you add `question_type` to the list of columns then each row will already have the question type and you will not need to do another query on line 165.

bnmnetp (changes requested):

controllers/assignments.py
bisterj: Are you talking about through the docstring starting at line 150 or somewhere else

bnmnetp: Line 150 is not a docstring.  It is a python string with an SQL query. -- a parameter to the `db.executesql` function on line 149.  And yes, this is what I was talking about.

bnmnetp: Looking more closely at the query I realized it is not as simple as just adding question_type to the select clause.  If a column is not in the group by clause you have to aggregate over it. In this case there is a one-to-one mapping from name to question type so we can use either min or max to aggregate and it will give us the same value.

this should do the trick...

```
select chapter, name, min(question_type) question_type, min(score), max(score), to_char(avg(score), '00.999') as  mean, count(score) 
from assignment_questions join questions on question_id = questions.id join question_grades on name =  div_id 
where assignment_id = %s and course_name = %s
group by chapter, name
```

Note that I used `min(question_type) question_type` so that the resulting column name will be called `question_type` and not `min`

bisterj: Thank you, that works better

bisterj (commented):
That file should be changed

bisterj:
@bnmnetp Fixed PR comments

PR #1810, opened 2021-11-18
rsmanage updates

bnmnetp:
- Use bookserver functions for DB initialization
- List deprecated functions in Ajax
- ignore the instantiated runestone


bjones1:
This is a great start on the process! I had a couple of ideas / design decisions to contribute -- I'll discus here instead of the newsgroup, since this seems closer to the problem.

Location
======
I propose moving `rsmanage` to the BookServer. This will help get us in the mindset of seeing the BookServer as a stand-alone program (PyPI-installable, runs on SQLite, etc.) Instructors need most of the `rsmanage` tools, and IMHO these should be provided in "stand-alone" mode without requiring installation of the web2py code.

Docker
======
I propose that `rsmanage` runs its commands inside the Docker container if the Runestone Dockerized application is running. For simplicity, `rsmanage` could attempt to import `ensure_in_docker` from `docker_tools_misc.py`. If that succeeds, call it before any `rsmanage` command. If the import fails, provide a dummy `ensure_in_docker` command that does nothing.

bjones1:
I'd suggest adding code like this to the beginning of `rsmanage`, so that it will automatically run itself in Docker.

``` Python
# Add the Docker tools to the Python path.
wd = (Path(__file__).parents[1]).resolve()
sys.path.extend([str(wd / "docker"), str(wd / "tests")])
# Run this in Docker if possible. We assume that a development version of the Runestone Server implies Docker.
try:
    from docker_tools_misc import ensure_in_docker
    ensure_in_docker()
except ModuleNotFoundError:
    pass
```

bnmnetp:
If the non-docker version of RunestoneServer is not a supported configuration should we be running CI on it?

bjones1:
Good point. My thoughts:

- CI works for now. Let's postpone this until after launch.
- After launch, we can run CI in Docker on the Runestone Server, the BookServer, and the Runstone Components by calling `docker-tools test`. But this will take longer to run and is more fragile (a single error in any of the three causes an overall test failure). What makes sense?

PR #1815, opened 2021-11-23
Development Cloning & Various small changes

Cortesj:
Adds 4 new flags to the **BUILD** command pertaining to repository cloning. There is checks to see if remote repository is a valid repository and will clone the default RunestoneInteractive remote repositories if a error such as 'git error 128 ' or any other subprocess error is passed into script. 

Fixes a issue with 1.py template located within docker_tools.py. There was a syntax error with the f string that uses RUNESTONE_PATH, import error of missing import for function ENVIRON and missing "_" from settings variable. 

Added to ```docker_tools.py build [options]```
- ```-crs / --clone-rs``` | Clone RunestoneServer with "X" Username
- ```-cbs / --clone-bs``` | Clone BookServer with "X" Username
- ```-crc / --clone-rc``` | Clone RunestoneComponents with "X" Username
- ```-c / --clone-all``` | Clone All repositories with "X" Username

Related to my issue #1775

Would @bjones1 and @bnmnetp  be able to review this to see if I need to potentially make some changes or potentially relocated to some other script.


docker/docker_tools.py
```diff
@@ -248,12 +276,23 @@ def build(
             print(f"Unable to run git: {e} Installing...")
             xqt("sudo apt-get install -y --no-install-recommends git")
 
+        # Check if Clone-All Flag is set
```
bjones1, line 279: Also check that the other clone flags aren't set. If they are, issue a warning that their values are ignored, or even quit with an error.

docker/docker_tools.py
```diff
@@ -248,12 +276,23 @@ def build(
...
+
         # Are we inside the Runestone repo?
         if not (wd / "nginx").is_dir():
             change_dir = True
             # No, we must be running from a downloaded script. Clone the runestone repo.
             print("Didn't find the runestone repo. Cloning...")
-            xqt("git clone https://github.com/RunestoneInteractive/RunestoneServer.git")
+            try:
+                # Check if possible to clone RunestoneServer with Custom Repo
+                xqt("export GIT_TERMINAL_PROMPT=0 && git clone https://github.com/%s/RunestoneServer.git"%clone_rs)
```
bjones1, line 292: You can use `xqt("GIT_TERMINAL_PROMPT=0...` instead (no need to export).

docker/docker_tools.py
```diff
@@ -248,12 +276,23 @@ def build(
...
+            except subprocess.CalledProcessError as e:
```
bjones1, line 293: I suggest omitting this -- if the repo clone fails, exit with an error. Otherwise, the user won't see the difference in all the output produced by a build.

docker/docker_tools.py
bjones1: Also, use an f-string instead -- that's easier to read.

docker/docker_tools.py
```diff
@@ -317,31 +358,52 @@ def build(
                     print(
                         f"Dev mode: since {bks} doesn't exist, cloning the BookServer..."
                     )
-                    xqt("git clone https://github.com/bnmnetp/BookServer.git")
+                    try:
+                        # Check if possible to clone BookServer with Custom Repo
+                        xqt("export GIT_TERMINAL_PROMPT=0 && git clone https://github.com/%s/BookServer.git"%clone_bs)
+                    except subprocess.CalledProcessError as e:
```
bjones1, line 364: Same comment -- exit with error.

docker/docker_tools.py
```diff
@@ -317,31 +358,52 @@ def build(
                     print(
                         f"Dev mode: since {bks} doesn't exist, cloning the BookServer..."
                     )
-                    xqt("git clone https://github.com/bnmnetp/BookServer.git")
+                    try:
+                        # Check if possible to clone BookServer with Custom Repo
+                        xqt("export GIT_TERMINAL_PROMPT=0 && git clone https://github.com/%s/BookServer.git"%clone_bs)
```
bjones1, line 363: Same comment -- no `export`.

docker/docker_tools.py
```diff
@@ -317,31 +358,52 @@ def build(
...
                 if not rsc.exists():
                     print(
                         f"Dev mode: since {rsc} doesn't exist, cloning the Runestone Components..."
                     )
-                    xqt(
-                        "git clone --branch peer_support https://github.com/RunestoneInteractive/RunestoneComponents.git"
-                    )
+                    try:
+                        # Check if possible to clone RunestoneComponents (peer_support) with Custom Repo
+                        xqt(
```
bjones1, line 374: Same comments.

docker/docker_tools.py
```diff
@@ -317,31 +358,52 @@ def build(
...
             docker_sudo = True
+            
 
         # Provide this script as a more convenient CLI.
         xqt(
             f"{sys.executable} -m pip install --user -e docker",
         )
 
-        # Run the Docker build.
+        # Filter out clone flags
```
bjones1, line 396: Why do this? I'd suggest keeping all the flags, just in case we want to refer to them later.

bjones1 (changes requested):

docker/docker_tools.py
Cortesj: I'm confused because default repositories is "RunestoneInteractive" and if the ```-c / --clone-all``` flag is even touched then it will set all the  repositories to attempt cloning with specified username. Are you saying that I should implement a check for each repository that will error out/exit if it cannot clone from specified user's remote repositories?

docker/docker_tools.py
Cortesj: The only reason I removed the clone flags from being passed though is due to issues with the actual building of the docker containers. I might be able to replicate the issue.

docker/docker_tools.py
```diff
@@ -184,8 +184,32 @@ def cli() -> None:
     default=False,
     help="Install tools needed for development with the PIC24/dsPIC33 family of microcontrollers.",
 )
+@click.option(
```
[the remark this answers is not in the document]
Cortesj: I never heard of black till now and it seems like such a sweet little program! I used it in latest commit to PR.

docker/docker_tools.py
Cortesj: Must have forgot f-string is a thing again... Anyway, I changed them to f-strings like you suggested and it appears that export is necessary to temporarily disable the Git Prompt to pass-though the error. I attempted to execute it without ```export``` but it would act as if it was never executed. I did noticed that the export only affects that session of script execution so it isn't exporting to the host system.

docker/docker_tools.py
Cortesj: I decided to change the except to perform a "sys.exit()" with a message stating there was an error cloning the system a remote repository with specific username.

docker/docker_tools.py
Cortesj: Changed exit to error out with message.

docker/docker_tools.py
Cortesj: Previously commented about ```export``` command being used for cloning.

docker/docker_tools.py
Cortesj: Noted and changed in latest Commit.

Cortesj:
Latest commit 8886946aa1cb8368d6708bf891b1827f88f8fcba I did the majority of the suggestions about system error exits, f-strings, and  etc. Waiting on some more suggestions and discussion about ```export``` and passing all arguments to the containers. I knew you had some questions about not passing the clone flags.

docker/docker_tools.py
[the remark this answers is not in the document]
bjones1: Great!

docker/docker_tools.py
bjones1: Here, I'm asking for a CLI consistency check: it doesn't make sense (to me) to specify both `--clone-all=my_repo` while also specifying `--clone_rc=my_other_repo`. I can see three solutions:

1. In the help text, state that value supplied in `--clone-rs`, `--clone-bs`, and `--clone-rc` will override the value supplied in `--clone-all`. Also issue a warning when this happens.
2. Issue an error, telling user that the same value was specified twice.

I'll leave it to you to decide which is best.

On your second point: yes, you definitely want to error if a git clone from the specified repo fails.

docker/docker_tools.py
bjones1: Good. I think that `sys.exit()` requires an int, though? You can use [click.echo](https://click.palletsprojects.com/en/8.0.x/quickstart/#echoing) to display a message.

docker/docker_tools.py
bjones1: OK, let me know.

bjones1 (changes requested):

docker/docker_tools.py
Cortesj: ```sys.exit()``` has an optional argument. If nothing is specified within the function or ```None``` value is passed then the default exit code is 0. I know you can pass a int between 0-127 to refer to different error types. I passed a string object which prints out to STDERR (Terminal) and returns an exit code 1. 

It worked during testing by providing it non-existing usernames/repositories to see if ERROR reported out correctly.

Reference:
[Python sys module Documentation](https://docs.python.org/3/library/sys.html)

docker/docker_tools.py
Cortesj: > Here, I'm asking for a CLI consistency check: it doesn't make sense (to me) to specify both --clone-all=my_repo while also specifying --clone_rc=my_other_repo. I can see three solutions:

Okay, I think I see your confusion. You are only meant to specify ```--clone-all``` if you wish to clone all the repositories with your remote repositories. Hypothetically, if you only want to clone BookServer from your own remote repository then you would specify ```--clone-bs```. 

I think I will be more specific within the ```--help``` stating the outcome of using ```--clone-all``` in tandem with any other clone flag. I could also provide some examples within ```--help``` if that would help with increasing the thought behind its functionality.
<br><br/>
> On your second point: yes, you definitely want to error if a git clone from the specified repo fails.

Addressed this with ```sys.exit()``` on another comment.

docker/docker_tools.py
bjones1: OK, sounds good.

docker/docker_tools.py
bjones1: Thanks, I learned something (sys.exit(string))!

bjones1:
This is close -- AFAIK, the only remaining items are:

1. Add clarification / help messages on what happens if the user specifies `--clone-all` and `--clone-xx`
2. Clean up the arg pass to Docker.

Cortesj:
Created a few new commits to address the suggestions involving cloning clarification and help messages

1. Commit 9e1c34ae7e9361b70d46ca267b4d471bbc08e98d
* Added a WARNING for using the ```--clone-all``` flag within help description
* Changed the METAVAR to display <USERNAME> rather than TEXT
Screenshot from 2021-11-24 00-15-54

* Removed the argument pass filter since after testing I could not replicate previous issue
* Initiate 20 second countdown on how to abort script if ```--clone-all``` flag is set
Screenshot from 2021-11-24 00-09-36

2. Commit e4e0bc4238cab6b235086374fef688117eca66ff
Clean up of code with ```Black```

docker/docker_tools.py
Cortesj: Addressed this with commit 9e1c34ae7e9361b70d46ca267b4d471bbc08e98d

docker/docker_tools.py
Cortesj: After some testing, I never was able to replicate the error. I currently tested it with multiple combinations of the ```--clone-XX``` flag.

docker/docker_tools.py
```diff
@@ -248,12 +286,39 @@ def build(
...
+        # If Clone-all flag is set override other Clone Flags
+        if clone_all != "RunestoneInteractive":
+            # Print Warning and provide countdown to abort script
+            click.secho(
+                "Warning: Clone-all flag was initalized and will override any other clone flag specifed!",
+                fg="red",
+            )
+            from time import sleep
+
+            # Wait 20 Seconds prior to proceeding with script
```
bjones1, line 298: I'd omit this. A warning in red is enough.

bjones1 (changes requested):
Mostly there! Just remove the 20 second delay.

docker/docker_tools.py
Cortesj: Commit 4bccb7c024d88799ceed7007694838722406baf1 removes delay.

bjones1 (approved):
Look good!

Cortesj:
@bjones1 It seems my pull request failed to merge due to the server tests. I know that within Docker_build the error comes to not setting the BookServer clone default to "bjones1". I am just confused on why ubunutu_build failed.

bjones1:
It looks like Github was down/not working well which caused the `ubuntu_build` failures. Just make a small change and push and it will re-run the tests. They should work better after a retry.

Cortesj:
@bjones1 Okay, I made a few small changes mostly some misplaced variable names and the default for BookServer since it isn't officially forked by RunestoneInteractive yet. It says there is a merge conflict only those with write access can fix but I am not sure why it got flagged as merge conflict since it simply replaces some current code for RunestoneComponents with what I wrote.

PR #1817, opened 2021-11-23
Fix: Avoid exceptions with new db constraints.

bjones1:
Avoid error discovered running with the new db constraints on my production machine.

### DIFF
--- a/models/db.py
+++ b/models/db.py
@@ -538,22 +538,17 @@ if "auth_user" in db:
 def admin_logger(logger):
     if settings.academy_mode:
         if auth.user:
-            sid = auth.user.username
-            course = auth.user.course_name
-        else:
-            sid = "Anonymous"
-            course = "Unknown"
-        try:
-            db.useinfo.insert(
-                sid=sid,
-                act=request.function,
-                div_id=request.env.query_string or "no params",
-                event=request.controller,
-                timestamp=datetime.datetime.utcnow(),
-                course_id=course,
-            )
-        except Exception as e:
-            logger.error(f"failed to insert log record for practice: {e}")
+            try:
+                db.useinfo.insert(
+                    sid=auth.user.username,
+                    act=request.function,
+                    div_id=request.env.query_string or "no params",
+                    event=request.controller,
+                    timestamp=datetime.datetime.utcnow(),
+                    course_id=auth.user.course_name,
+                )
+            except Exception as e:
+                logger.error(f"failed to insert admin log record: {e}")
 
 
 def createUser(username, password, fname, lname, email, course_name, instructor=False):


bjones1:
Is this the correct approach? I'm basically discarding any logging if a user isn't logged in, since we don't know what course or user to log.

models/db.py
```diff
@@ -538,22 +538,17 @@ def check_for_donate_or_build(field_dict, id_of_insert):
 def admin_logger(logger):
     if settings.academy_mode:
         if auth.user:
-            sid = auth.user.username
-            course = auth.user.course_name
-        else:
```
bnmnetp, line 543: I don't understand this change.

What was causing the exceptions that this PR is trying to avoid?

bjones1: This attempts to insert a `useinfo` record where `useinfo.course_id == "Unknown"`. However, since `useinfo.course_id` [is a foreign key](https://github.com/bnmnetp/BookServer/blob/main/bookserver/models.py#L149) for `courses.course_name`, this violates the constraint: the course name "Unknown" does not exist in the `courses` table.

bnmnetp: The foreign key constraint is new.  I would rather use "boguscourse" which has always been a placeholder course than to simply lose all of this information.  This is valuable information for analyzing the use of the admin interface.

bjones1: That sounds good to me. Currently, the BookServer doesn't create `boguscourse` in `crud.create_initial_courses_users`. I assume we'll remove this function, and instead use `rsmanage initdb` instead? (I see `boguscourse` in `rsmanage.initialize_tables`).

bjones1:
Finding this bug brings up another issue: an anti-pattern scattered throughout the Runestone code base, usually in the context of logging. The exception handler leaves the database in a broken state, meaning that all past and future commits will be discarded, producing useless traceback info after the request completes. I propose not catching the error, to provide immediate feedback instead.

The pattern is:

``` Python
try:
    # Some db operation: for example,
    db.useinfo.insert(stuff_that_breaks)
except Exception as e:
    logger.error(f"failed to insert admin log record: {e}")
```

The resulting traceback is:

```
  File "/home/www-data/web2py/gluon/main.py", line 439, in wsgibase
    serve_controller(request, response, session)
  File "/home/www-data/web2py/gluon/main.py", line 202, in serve_controller
    raise HTTP(response.status, page, **response.headers)
gluon.http.HTTP: 200 OK

  File "/home/www-data/web2py/gluon/main.py", line 455, in wsgibase
    session._try_store_in_db(request, response)
  File "/home/www-data/web2py/gluon/globals.py", line 1254, in _try_store_in_db
    record_id = table.insert(**dd)
  File "/home/www-data/web2py/gluon/packages/dal/pydal/objects.py", line 891, in insert
    ret = self._db._adapter.insert(self, row.op_values())
  File "/home/www-data/web2py/gluon/packages/dal/pydal/adapters/base.py", line 546, in insert
    raise e
  File "/home/www-data/web2py/gluon/packages/dal/pydal/adapters/base.py", line 541, in insert
    self.execute(query)
  File "/home/www-data/web2py/gluon/packages/dal/pydal/adapters/__init__.py", line 69, in wrap
    return f(*args, **kwargs)
  File "/home/www-data/web2py/gluon/packages/dal/pydal/adapters/base.py", line 468, in execute
    rv = self.cursor.execute(command, *args[1:], **kwargs)
psycopg2.errors.InFailedSqlTransaction: current transaction is aborted, commands ignored until end of transaction block
```

This gives absolutely no clue where the error occurred. I found it by browsing my server logs, which was not an obvious approach to me.

Discussion
=======
There are two cases for this error:
1. There were commits before the exception occurred. Then we should raise an error immediately, since we're losing data.
2. There were no commits before the exception occurred. Then we should do a rollback and begin another transaction, which (I think) would allow everything to proceed normally.

However, there's no way I know of to distinguish these two cases. A third option: use some Postgres-specific code (a savepoint?) to run. But this means we won't support sqlite.

Therefore, the simplest approach is to raise an error immediately. We should also search our logs to see if these errors are happening elsewhere and work around them. Here's what I found in my logs:

```
uwsgi.log:ERROR:web2py.app.runestone:failed to insert log record for practice: insert or update on table "useinfo" violates foreign key constraint "useinfo_course_id_fkey"
uwsgi.log:ERROR:web2py.app.runestone:failed to insert log record for practice: insert or update on table "useinfo" violates foreign key constraint "useinfo_course_id_fkey"
uwsgi.log:ERROR:web2py.app.runestone:failed to insert log record for practice: insert or update on table "useinfo" violates foreign key constraint "useinfo_course_id_fkey"
uwsgi.log:ERROR:web2py.app.runestone:failed to insert log record for practice: insert or update on table "useinfo" violates foreign key constraint "useinfo_course_id_fkey"
uwsgi.log:ERROR:web2py.app.runestone:failed to insert log record for practice: insert or update on table "useinfo" violates foreign key constraint "useinfo_course_id_fkey"
uwsgi.log.1631735760:ERROR:web2py.app.runestone:failed to insert log record for practice: value too long for type character varying(512)
```

bnmnetp:
Good points.  I think in many cases though we do want to catch the exception rather than just have the end user seeing a stack trace.  That is not a good user experience.   But leaving the DB in a bad state is also not good.

As I understand it web2py automatically starts a transaction for each request and then commits it at the end unless there is an exception, then it is rolled back.   But if we are explicitly catching that exception then we are foiling web2py's plans.  I think in those cases it is up to us to do the right thing and rollback or attempt to recover in some other way.

I think there are only a few cases where we call commit ourselves. So perhaps it is not as hard to know if there has been a prior commit as you think. This does not feel like an emergency to me since we've been doing this for years, but we should definitely find a way to improve the situation.

I do occasionally get a giant streak of those `current transaction is aborted, commands ignored until end of transaction block` messages that appear in the errors folder. Presumably a side effect of what you described above.

bnmnetp:
I've modified my rsmanage_updates branch to use boguscourse. I also updated the code in book server to add boguscourse to the list.

We should create an issue to further discussion the problems you have identified above as they are really a separate issue from using a nonexistent course.

bjones1:
Sounds good!

Issue #1763, opened 2021-09-21
Modifying a grade from grade book drill-down does not update totals


bnmnetp:
From the grade book drill down into an assignment and question.  
Update the grade for that questions.

Going back to the grade book you will see that the overall grade is not updated.

When updating a grade from the drill-down panel, the total for the assignment should be updated. `do_update_total` should be called after the new grade is entered.

areskeldi:
@sreynit02 and I, Berea College students, are working on this issue

sreynit02:
@bnmnetp Which type of questions didn't update the total? We're having trouble recreating this issue.

bnmnetp:
All types of questions.

Make sure you are repeating this by going through the **gradebook** NOT the grading tab.

sreynit02:
Hello, It seems like `do_update_total` is not a function in RunestoneServer, but we found a function in rs_grading.py named `do_calculate_totals` that seem to compute the total score a student earns for an assignment. Did you mean this function? 

Also, if that's the case, we are not sure where to call this function because the gradebook gets data from grades() in dashboard.py which already has a logic that recalculates the total point using another function calls `update_total_points` .

PR #1825, opened 2021-12-02
solve issue #1763 (update total point in gradebook dashboard)

areskeldi:
@sreynit02 and I solved the issue #1763 making the following changes:
1. Pass assignment id with POST data object in admin.js
2. Validate new "assignmentid" parameter for the record_grade function
3. Gather assignment information
4. Call do_calculate_totals() in the record_grade function

This pull request was approved by Dr. Jan Pearce of Berea College.

**Screenshot walkthrough of the change**
The gradebook before instructor changes grade for assignment question:

image

Question drilldown:

image

Change point for first question from 0 to 1:

image

Point is updated in the question drilldown:

image

Now, the change to question points is also reflected in the gradebook (used to be 50%, now it's 100%, as expected):

image

### DIFF
--- a/controllers/assignments.py
+++ b/controllers/assignments.py
@@ -342,6 +342,10 @@ def record_grade():
         return json.dumps(
             {"success": False, "message": "Must provide either grade or comment."}
         )
+    if ("assignmentid" not in request.vars):
+        return json.dumps(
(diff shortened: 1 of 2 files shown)


bnmnetp:
fixes #1763

PR #1832, opened 2021-12-16
Add Custom Textbook Creation to Designer Page

cassid1:
- High Level Overview
  - This pull request aims to add Github integrated custom textbook editing and viewing by a complete overhaul of the "designer" controller to include both the old course creation page as well as the new textbook creation page. We also added functionality to the books controller to allow custom textbooks to be viewed by their owner even if they are not associated with a course. The custom textbooks are located in a new directory on the runestone server: `custom_books/<published or drafts>/<runestone username>/<book name>` and there is a new table in the database named "textbooks" to keep track of each custom one.
- Description of main functionality
  - **Verifying** The Github integration process began with Dr. Brad Miller registering a Github OAuth Application using the RunestoneInteractive Github account. This registration takes a callback URL and produces a client ID and client secret, both of which are stored as session variables. From our webpage, users click on a link to `github.com/login/oauth/authorize` and that combined with the client ID we send causes Github to ask them to grant access to the OAuth Application. Upon authorizing our app to edit their repositories, github redirects them to the callback URL we provided along with a code. Our callback function then sends this code, the client ID, and the client secret, back to github, and in turn we receive a token which will be used by all API calls. This token is stored as a session variable for security purposes, and it is constantly being validated and refreshed as needed behind the scenes after a user has granted runestone this permission. This method of Github authentication is much cleaner and more secure than any other approach, since we do not need to collect and store user’s actual Github credentials and instead are simply providing a link to an external login which in turn redirects back to us.
  - **Creating** First `github_book_repo` is called, and once it is verified that the repo does not already exist, the original book is forked from RunestoneInteractive to the user’s github account. Additionally, a webhook is created in that repository that will send data back to the runestone server. That webhook is given a copy of a randomly generated secret key, another copy of which is stored in the database, along with the fact that the book is not yet published.
  - **Updating** When the repo is forked (and also whenever there is a commit to the default branch), Github sends the Runestone server a commit id, which is passed to `update_local_files`, which will try pulling from github and setting the local repository to the version given by the commit id. If anything at all goes wrong, it will simply delete the local copy and re-clone it, a useful feature if the name or default branch of the Git repository changes. Additionally, before any of these changes happen, it is verified that the header sent by Github containing a sha256 HMAC of the webook secret and the json body content matches the corresponding hash generated using the same algorithm but with the secret key put in the database when the book was first created. 
  - **Viewing** Previously all books on the runestone server had been located in `books/`, however now there are custom versions in `custom_books/<drafts or published>/<username>/<book name>/` that also need to be served. Also, viewing a book from the book directory had always necessitated that the user was currently in a course using that book. For custom books we would like the owner of that custom book to be able to view both draft and published versions without having to make courses for them, while at the same time still be able to create a normal course for the published version of a custom book. This meant rerouting some of the book viewing to a new courseless way while keeping all of the old book viewing functionality.
  - **Publishing Editing and Deleting** Only the draft version of books are updated by github webhooks. Clicking “publish” will `update_local_files` with the current commit id for the draft book, making them match. If a custom book’s name is changed, the “textbooks” table will be updated along with the “courses” table, provided that there are any courses created using that book. In the case of deletion, courses using the now non-existent textbook will be set to instead use the book that it was originally cloned from, keeping any courses from breaking entirely. Editing and deleting also `update_local_files` to rename or remove them. In the case that the book’s github repo is edited, `github_book_repo` will make an API request to update the name. When a book is deleted, the corresponding repository will be as well.
- Issues
  - Had to remove database caching from route_book because sometimes the base_course of a course would change (this happens whenever the name of a custom book is changed...regular books don't experience this) and then it would be trying to get a bad base_course
  - We havent found a way to specify a repository name when forking a runestone book, this means the user cannot have a github repo of that name that already exists, or else the fork would fail. This also means that it doesn't really work for a user to have 2 different custom books at the same time that both come from the same base book
-  Possible expansions
  - Separate system for linking an already forked repository. These repositories would probably not be deleted if their runestone textbook was. This would also allow for our OAuth app to ask for fewer permisisons. As it stands we need to ask for fiull repository access and ability to delete all repositories


bnmnetp:
I'm working on getting this going at dev.runestoneacademy.org and running into some problems.  I created a new client id and secret and have logged into github, it returns me to the page like it ought to after I log in.  But when I try to clone a book I get one of three messages:

* Created book: notused
* trying to clone a repo that already exists. (OK, your note above explained this one)
* timeout waiting to be forked

None of them ever show up under my custom textbooks

I also ran into a problem because my test user did not have a folder under custom_books/published or custom_books/drafts

I hacked around that one easily, but I’m stuck right now as the clone is not working for me.

PR #1835, opened 2021-12-23
Fix: Get more unit tests passing.

bjones1:
- Provide more flexible test runner.
- Fix install problems with pgcli outside a container.


bnmnetp:
Shall I merge this so that I can merge bookserver#89 ?

PR #1862, opened 2022-01-11
Clean: De-duplicate settings in both 0.py and 1.py.prototype.

bjones1:
Put confidential settings in .env.prototype.


docker-compose.yml
```diff
@@ -37,6 +37,18 @@ services:
...
       LOAD_BALANCER_HOST: '${LOAD_BALANCER_HOST}'
+      # Stripe settings.
+      STRIPE_PUBLISHABLE_KEY: ${STRIPE_PUBLISHABLE_KEY}
+      STRIPE_SECRET_KEY: ${STRIPE_SECRET_KEY}
+      STRIPE_DEV_PUBLISHABLE_KEY: ${STRIPE_DEV_PUBLISHABLE_KEY}
+      STRIPE_DEV_SECRET_KEY: ${STRIPE_DEV_SECRET_KEY}
+      STRIPE_TEST_PUBLISHABLE_KEY: ${STRIPE_TEST_PUBLISHABLE_KEY}
+      STRIPE_TEST_SECRET_KEY: ${STRIPE_TEST_SECRET_KEY}
+      # E-mail settings.
+      EMAIL_SERVER: ${EMAIL_SERVER}
```
bnmnetp, line 48: Are these email env vars necessary?  If we define them we ought to use them in 1.py maybe?  I know it makes them available for the new bookserver as well, but we are not doing anything with email there yet.

bjones1: I'm trying to put all secrets in one place -- `.env`. The e-mail config includes a password, so I put all e-mail stuff together there.

I will update `1.py.prototype` to make use of these -- thanks! I somehow overlooked that.

bjones1: I just pushed fixes.

bjones1: This now contains the fix for the LTI bug. (I forgot an import -- sigh.)

bnmnetp: How in the world could this have been working for me all week without this!!??!!

bjones1: :)

bjones1: My earlier LTI "fix" really broke everything, sigh. But since the exception is caught and logged, it's not as obvious.

bnmnetp: Because we love how web2py handles imports! :-(

bnmnetp (approved):
I'm not sure what a better solution would be, but changing the names of things is really painful when you have a working 1.py that you have configured for your needs in a production system.

I've reviewed this carefully and hopefully updated my 1.py in all the right places but I just have to cross my fingers and be ready to fix whatever I may have missed quickly when I bring this up.

bjones1:
Thanks. I know how painful that is, and I really appreciate your work in doing these updates. I think it will be a good step for future admins who want to run their own server.

PR #1869, opened 2022-01-15
Fix: Make database name configurable.

bjones1:
Works in production.


bnmnetp (commented):
What is the point of this name change?  I don't see that it is fixing any bugs, its just stirring the pot and forcing me to update my existing (and working) .env files. ??

bjones1:
This supports a `docker-tools build --single` production configuration, where the database runs on a separate machine (my production setup). I agree that it definitely requires some reconfig for you.

bnmnetp (approved):

PR #1870, opened 2022-01-15
Fix: Remove backport repo after install.

bjones1:
Clean: Remove unused vars, imports.

Works in production.

### DIFF
--- a/docker/docker_tools.py
+++ b/docker/docker_tools.py
@@ -155,7 +155,6 @@ except ImportError:
 try:
     from docker_tools_misc import (
         add_commands,
-        get_bookserver_path,
         get_ready_file,
         in_docker,
         SERVER_START_FAILURE_MESSAGE,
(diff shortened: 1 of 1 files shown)


docker/docker_tools.py
```diff
@@ -813,8 +814,6 @@ def _build_phase_2_core(
     exec(open(activate_this_path).read(), {"__file__": activate_this_path})
 
     w2p_parent = Path(env.WEB2PY_PATH).parent
-    bookserver_path = get_bookserver_path()
```
bnmnetp, line 816: This does not seem related to the title of this PR, and not getting the bookserver_path seems concerning. What is this about?

bjones1: This variable is never used -- I moved it to `docker_tools_misc.py`. This is cleanup from flake8.

bjones1: (By never used, I mean that it's used only in the next line; `run_bookserver_kwargs` isn't used, making `bookserver_path` also unused).

PR #1872, opened 2022-01-15
Fix: Clean up flags for rsmanage addcourse.

bjones1:
Works in production.

### DIFF
--- a/rsmanage/rsmanage.py
+++ b/rsmanage/rsmanage.py
@@ -280,10 +280,9 @@ def shutdown(config):
 @click.option(
     "--start-date", default="2001-01-01", help="Start Date for the course in YYYY-MM-DD"
 )
-@click.option("--python3", is_flag=True, default=True, help="Use python3 style syntax")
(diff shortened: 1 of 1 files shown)


rsmanage/rsmanage.py
```diff
@@ -280,21 +280,19 @@ def shutdown(config):
 @click.option(
     "--start-date", default="2001-01-01", help="Start Date for the course in YYYY-MM-DD"
 )
-@click.option("--python3", is_flag=True, default=True, help="Use python3 style syntax")
```
bnmnetp, line 283: I'm Not sure what the difference is between old and new, I thought that is_flag automatically implied the --no version of the option.  But maybe this is more clear/explicit.

bjones1: Per the [click docs](https://click.palletsprojects.com/en/8.0.x/options/#boolean-flags), the old code creates an option without the `--no` version of the option:

> If you really don’t want an off-switch, you can just define one and manually inform Click that something is a flag:

```
import sys

@click.command()
@click.option('--shout', is_flag=True)
```

PR #1874, opened 2022-01-15
Fix: Avoid printing password to stdout.

bjones1:
Works in production.

### DIFF
--- a/docker/docker_tools.py
+++ b/docker/docker_tools.py
@@ -864,7 +864,7 @@ def _build_phase_2_core(
             {sys.executable} -c \\
             "from gluon.main import save_password
             save_password(
-                '{env.WEB2PY_ADMIN_PASSWORD}',
+                '$WEB2PY_ADMIN_PASSWORD',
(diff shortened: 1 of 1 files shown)


docker/docker_tools.py
```diff
@@ -864,7 +864,7 @@ def _build_phase_2_core(
             {sys.executable} -c \\
             "from gluon.main import save_password
             save_password(
-                '{env.WEB2PY_ADMIN_PASSWORD}',
```
bnmnetp, line 867: I do not understand this at all... Seems like this changes to the password to the literal string `$WEB2PY_ADMIN_PASSWORD`

bjones1: This is being run via a bash shell -- the full command after interpreting the string and assuming we're using HTTP is:

```
python3 -c 
"from gluon.main import save_password
save_password('$WEB2PY_ADMIN_PASSWORD', 80)"
```

Bash will replace `$WEB2PY_ADMIN_PASSWORD` with its actual value, while still not printing that value to the log.

PR #1878, opened 2022-01-16
Fix: Install correct version of qemu.

bjones1:


Dockerfile
```diff
@@ -7,7 +7,7 @@
 # ************************************************************
 # To build, execute `docker-tools <docker/docker_tools.py>` ``build``.
 
-FROM python:3.9-buster
```
bnmnetp, line 10: Nice that we can upgrade a whole operating system with just a couple of characters.

bjones1: Amazing! Even better, the downtime is so minimal -- I build another container, stop the current container, then re-start the new container.

docker/docker_tools.py
```diff
@@ -501,21 +505,6 @@ def _build_phase_0(
         )
 
     if build_config.is_dev():
-        # OS X doesn't need this.
-        if is_linux:
-            # Poetry requires an executable named ``python`` -- make sure it's installed.
```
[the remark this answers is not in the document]
bjones1: This is the install for Poetry outside the container (on the host OS). We still [install Poetry inside the container](https://github.com/RunestoneInteractive/RunestoneServer/pull/1878/files#diff-abb0119db171f8a953001aa7bd9283957d1aa59d784658a89f80c8bde8a2b678R742).

PR #1898, opened 2022-02-22
Clean: Black.

bjones1:


bnmnetp:
Just to be clear, you are running black 22.1.x ??  22.1 is supposed to be stable for the year so we don't have to worry about competing versions of black.

bjones1:
Yes, correct:

```
root@b17b58543016:/srv/web2py/applications/runestone# black --version
black, 22.1.0 (compiled: yes)
```

PR #1899, opened 2022-02-22
Fix: Skip failing tests that don't make sense or need to be ported.

bjones1:


tests/test_server.py
```diff
@@ -1138,6 +1234,9 @@ def test_pageprogress(test_client, runestone_db_tools, test_user_1):
     assert '"subc_b_fitb": 0' in test_user_1.test_client.text
 
 
+@pytest.mark.skip(
+    reason="Can't render new BookServer template using old server. What does this test do? Should it be ported?"
```
bnmnetp, line 1238: lockdown mode is what the michigan people use in their instances that they use with coursera.

bnmnetp (approved):

PR #1900, opened 2022-02-22
Clean: Remove duplicate ID.

bjones1:
VS Code beautify of HTML.

### DIFF
--- a/views/default/user.html
+++ b/views/default/user.html
@@ -32,9 +32,32 @@
             {{else: }}
                 <div class="carousel slide col-md-12" data-ride="carousel" id="myCarousel"  style="margin-bottom: 50px;">
             {{ pass }}
-            <p style="text-align: center;">
-                <img src="/static/images/logo_small.png" width="400" style="margin: auto;" >
-            </p>
+                <!-- Indicators -->
+                <ol class="carousel-indicators">
+                    <li data-target="#myCarousel" data-slide-to="0" class="active"></li>
+                    <li data-target="#myCarousel" data-slide-to="1"></li>
+                </ol>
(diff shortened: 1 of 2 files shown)


views/default/user.html
```diff
@@ -32,9 +32,32 @@
             {{else: }}
                 <div class="carousel slide col-md-12" data-ride="carousel" id="myCarousel"  style="margin-bottom: 50px;">
             {{ pass }}
-            <p style="text-align: center;">
-                <img src="/static/images/logo_small.png" width="400" style="margin: auto;" >
-            </p>
+                <!-- Indicators -->
```
bnmnetp, line 35: This seems to undo my earlier changes where I removed the carousel and updated outdated links.

bnmnetp (changes requested):

bnmnetp:
I'm going to close this as I was working on some updates to the books/index.html
I eliminated the duplicated id in the book list.

I am curious what happened with user.html as it was trying to put back things I removed and updated a few weeks ago.

bnmnetp:
I assume you are using prettier?   It seems to be fine in this case, but sometimes it does really bad things with the template blocks.

bjones1:
Oops, my mistake -- looks like my branch has outdated code. Thanks re: fixing this when you update `books/index.html`.

bjones1:
Yes, I'm using prettier. I agree that it doesn't play well with templates. Sigh -- I like that it does clean up the HTML code, but don't like its action on templates.

bnmnetp:
I pushed changes to `books/index.html` an hour ago

bjones1:
Got it, thanks!

PR #1906, opened 2022-02-27
DRY: Call a common function to create a JWT.

bjones1:


controllers/assignments.py
```diff
@@ -595,9 +595,7 @@ def doAssignment():
         # this means the user is logged in to web2py but not fastapi - this is not good
         # as the javascript in the questions assumes the new server and a token.
         logger.error(f"Missing Access Token: {auth.user.username} adding one Now")
-        _create_access_token(
-            {"sub": auth.user.username}, expires=datetime.timedelta(days=30)
-        )
+        create_rs_token()
```
bnmnetp, line 598: This just seems like you are wrapping the function that is already there.  That does not seem any DRYer to me, what am I missing?

bjones1: This at least takes the parameters (time to expire, user name) out of a bunch of calls and puts it in one single function.

It's also a step towards adding to the JWT -- future PRs will add to the contents, and putting this in a single function is much better than duplicating it 8 times or so.

bnmnetp: What kinds of things do you anticipate adding to the JWT ???

bjones1: [Scopes](https://fastapi-login.readthedocs.io/advanced_usage/#oauth2-scopes) (in the OAuth2/JWT sense) -- I'm using this for high-stakes assessments (exams) to allow access to only the exam part of the book or to the book except for the exam.

Issue #1944, opened 2022-05-29
Docker setup issues on macos


bnmnetp:
A user reported the following issues when trying to get a new Docker setup running on macOS.

1.  usermod:
on line 541 of docker_tools.py, there is a usermod shell command, but it seems that usermod is not a command that natively exists on mac. Instead, I manually had to do some combination of the following:

```
sudo dscl . create /Groups/www-data
sudo dseditgroup -o edit -a $USER -t user www-data
sudo dscl . append /Groups/www-data GroupMembership $USER
```

(I think that I somehow ended up needing both dscl and dseditgroup, because mac's group membership logic seemed weird and the 'groups' command didn't always reflect the new membership until I did both)

2. Xvfb:
I was building with --single-dev since I wanted to have RunestoneComponents and BookServer built from source, since I will probably end up making changes in at least one of those.
docker_tools.py tried to set up VNC access using Xvfb around line 817.

While Xvfb does exist, it didn't like the options that the script gave it. It failed with:
Unrecognized option: /private/tmp/com.apple.launchd.mkm3ilvZCX/org.xquartz:0
(I think the xquartz location was correct, but Xvfb didn't know what to do with that)

This particular failure didn't cause the script to crash, so I ended up just letting it happen because I still wasn't sure what options my Xvfb did understand.

3. chgrp with bad symlinks (?):
on line 888, docker_tools.py tries to run chgrp to recursively change permissions. I consistently got the following error:
chgrp: changing group of '/srv/web2py/applications/runestone/.venv/bin/python3': No such file or directory
(plus a few more for other versions of python)
the most likely cause seems to be that these are symlinks that aren't pointing to anything (or pointing to something chgrp can't deal with?). I'm not sure why they are there, since this is already inside the (fresh) docker image as far as I can tell. I tried a couple of options to convince chgrp to not follow symlinks, but it didn't seem to help. In the end I just forced the QUICK_START option to avoid this logic.

4. When restarting the script after various failures, I had a couple of issues with the logic starting on line 402:

first, I tried to re-run the script right after manually fixing the group permissions stuff; line 402 correctly detected that I wasn't yet inside the RunestonServer folder, but then it incorrectly said "Didn't find the runestone repo. Cloning...", and failed, because in fact the RunestoneServer folder already existed, because the previous run of the script failed after cloning the RunestoneServer repo.

[...]

PR #1945, opened 2022-06-09
OS X fix

bjones1:
Fix: Corrections on OS X courtesy of @yanamal. See #1944.

### DIFF
--- a/docker/docker_tools.py
+++ b/docker/docker_tools.py
@@ -70,7 +70,12 @@ if sys.platform == "win32":
     sys.exit()
 
 # See if we're root.
-is_root = subprocess.run(["id", "-u"], capture_output=True, text=True, check=True).stdout.strip() == "0"
+is_root = (
+    subprocess.run(
+        ["id", "-u"], capture_output=True, text=True, check=True
+    ).stdout.strip()
+    == "0"
+)
 
 
 # Check to see if a program is installed; if not, install it.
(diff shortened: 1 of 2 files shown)


bnmnetp:
Is this ready?

yanamal:
I just tested it by resetting my group membership (hopefully) back to the way it was before Runestone and re-running the script.

It works, with one caveat: because of sudo, it stops everything and asks for my password in the terminal when it reaches that point. Not sure if this is bad for users who might not expect it? (e.g. they start the script and then switch tabs). I think it does need sudo privileges to do the commands, but I'm not 100% sure of that either.

bjones1:
@yanamal, thanks for testing! That's typical for this script -- it will need sudo permission in most cases, and will (hopefully) ask for it almost immediately after starting, so that users won't have switched away from that window.

bjones1:
@bnmnetp, do you want to review this before merge, given your a Mac person?

docker/docker_tools.py
```diff
@@ -87,8 +92,8 @@ def check_install(
         print("Not found. Installing...")
         subprocess.run(
             # Only run with ``sudo`` if we're not root.
-            ([] if is_root else ["sudo"]) +
-            [
+            ([] if is_root else ["sudo"])
+            + [
                 "apt-get",
                 "install",
                 "-y",
```
bnmnetp, line 99: I think this is missing the install_package

PR #1957, opened 2022-06-27
Fix: Restart all containers on startup/crash.

bjones1:


bjones1:
@bnmnetp, this is per our conversation on Slack. It should also make use a bit simpler for users, since users no longer need to manually start the containers.

bnmnetp (approved):

PR #1961, opened 2022-06-28
Add: Send Celery log data to the Docker logs.

bjones1:
Docs: Prettify a bit.

### DIFF
--- a/docker/docker_tools_misc.py
+++ b/docker/docker_tools_misc.py
@@ -88,7 +88,9 @@ def _start_servers(dev: bool) -> None:
     xqt(
         'sudo -u www-data env "PATH=$PATH" "REDIS_URI=$REDIS_URI" '
         "poetry run celery --app=scheduled_builder worker --pool=threads "
-        "--concurrency=3 --loglevel=info --logfile=/dev/stdout &",
+        "--concurrency=3 --loglevel=info "
+        # This redirect ensures output ends up in the Docker log even if the servers are restarted. Sending to ``/dev/stdout`` only works at initial startup, but doesn't redirect after the servers are restarted.
+        "--logfile=/proc/1/fd/1 &",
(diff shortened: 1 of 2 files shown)


docker/docker_tools_misc.py
```diff
@@ -88,7 +88,7 @@ def _start_servers(dev: bool) -> None:
     xqt(
         'sudo -u www-data env "PATH=$PATH" "REDIS_URI=$REDIS_URI" '
         "poetry run celery --app=scheduled_builder worker --pool=threads "
-        "--concurrency=3 --loglevel=info &",
+        "--concurrency=3 --loglevel=info --logfile=/dev/stdout &",
```
bnmnetp, line 91: celery probably will not get restarted very often but the redirect I used below `/proc/1/fd/1` ensures that the log output continues to go to the docker log even after the server is restarted.  Whereas a simple redirect to stdout does not.

bjones1: Thanks! I'll update that. Do you have any docs / links to discuss why this works? I'd like to include that if so. It's a neat technique.

bnmnetp: I found it while searching for a solution.  It is discussed in several places.  Such as https://github.com/moby/moby/issues/19616 and others.

bjones1: Great, thanks! I'll add that link.

bjones1:
Ignore the test failure -- that's Poetry complaining that dependencies are outdated. The latest check ran on a docs-only commit; the previous check (on code changes) passed.

bjones1:
Whoops, I realized I committed changes to `pyproject.toml` in this PR -- I meant that to be in a separate PR...

bnmnetp:
I think thats fine, we wanted that in there anywhere.

PR #1970, opened 2022-07-01
Move: Put Celery code in the BookServer.

bjones1:
Stop supporting old web2py-based LP problems as part of the move.


bjones1:
This PR requires https://github.com/RunestoneInteractive/BookServer/pull/102 to be merged first in order to pass.

bnmnetp:
I merged BookServer#102 but this still fails...  

There is an error message about a path `None/bookserver` which seems pretty clearly wrong.

pyproject.toml
```diff
@@ -25,7 +25,6 @@ altair = "^4.0.0"
 beautifulsoup4 = "^4.0.0"
 bleach = "> 3.1.1"
 bookserver = "^1.0.0"
-celery = { extras = ["redis,gevent"], version = "^5.0.0" }
```
bnmnetp, line 28: Shouldn't there be a poetry.lock file corresponding to this change to pyproject.toml ??

bjones1:
Fixed now.

pyproject.toml
bjones1: I added this -- thanks.

bjones1:
This is a larger change, following up on PRs in the BookServer to fix how Celery works. I don't see easy ways to split this into smaller PRs.

bnmnetp (approved):

PR #1972, opened 2022-07-05
Docs: Improvements after working with Barb to install the servers.

bjones1:


docker/README.rst
```diff
@@ -60,6 +60,10 @@ If running on OS X, install then run the `Docker Desktop <https://www.docker.com
 
         Expect to wait a few minutes the first time you start the Docker Desktop. Don't proceed until the Docker Desktop's initialization is complete.
 
+        On OS X 12.4, Apple's new security features require you to give Docker `additional permissions <https://support.apple.com/guide/mac-help/control-access-to-files-and-folders-on-mac-mchld5a35146/mac>`_ if you place the files you install below in the Documents folder.
```
bnmnetp, line 63: We should recommend against installing anything in the Documents folder.  For most mac users this is a folder that is synced to iCloud and is not really an appropriate place for doing development work.

docker/README.rst
```diff
@@ -3,7 +3,7 @@ Docker Deployment
 
 .. TODO
 
-    Docs to update:
+    Is there some way to simplify these instructions? It's still a long, error-prone process! Do we need another setup script that follows these directions and prompts the user for input (do you want to run the server or do development? What is your github userid? Etc.)? Such a script would be able to take care of the first 7 steps, which would be good progress. But, we'd need os-dependent stuff for the first bits (install WSL/Ubuntu on Windows); can we assume Python 3 on Mac?
```
bnmnetp, line 6: I like the idea of a script that can automate even more of this.
  It would allow us to autodetect if they are on windows or linux or whatever.

I wonder if there isn't some kind of command we can have them do at the end of each step to ensure that the step was successful.

docker/README.rst
bnmnetp: I would have them create a subdirectory of their home directory called Runestone So that everything we installed gets installed under Runestone.

docker/README.rst
bnmnetp: There is not much we can do other than remind people that Docker on macOS and Windows is not as rock solid as it is on Linux.  Sometimes you just have to reboot.  Its not the fault of Runestone its a flaw in Docker.  -- Which for me has been incredibly stable for the last several weeks...  They have come a long way in get docker stable on the Apple Silicon.  I guess the advice should be that if docker run hello-world does not work then you need to restart the docker desktop.

docker/README.rst
```diff
@@ -89,15 +93,51 @@ Next, install ``curl`` by opening an Ubuntu terminal then typing:
...
+
+        python3 -m venv runestone
+
+#.  `Activate the virtual environment <https://packaging.python.org/en/latest/guides/installing-using-pip-and-virtual-environments/#activating-a-virtual-environment>`_ you just created.
+
+    On Windows:
+
+        .. code-block:: bash
+
+            runestone\Scripts\activate
```
bnmnetp, line 130: maybe rsve is a more descriptive, less confusing name than putting the runestone virtual environment in runestone especially when there are so many other folder names that are either runestone or start with runestone?

bjones1: Good point -- let me think of a more descriptive name...

docker/README.rst
bjones1: Resolving for now -- these are longer-term goals than just this PR.

docker/README.rst
bjones1: Added -- I call the directory `rsdocker` to be a bit more descriptive.

docker/README.rst
bjones1: I'm using `rsvenv` instead.

PR #1975, opened 2022-07-10
Fix: In poetry_fix, ignore missing poetry.lock

bjones1:
Remove unused docs
Docs: Improve.


bjones1:
This is a small change, and shouldn't require much time to review.

PR #1976, opened 2022-07-10
Clean: Collect version numbers in one place in docker-tools script.

bjones1:


bjones1:
This is a small change, and shouldn't require much time to review.

bnmnetp:
Good idea!

bjones1:
Thanks!

PR #1977, opened 2022-07-10
Fix: Simplify the user-facing install process

bjones1:
This puts all the cloned repos in the `rsdocker/` subdirectory.

On OS X, it runs everything in a venv placed in `rsdocker/rsvenv`.

There are also lots of docs fixes and improvements.

### DIFF
--- a/docker/README.rst
+++ b/docker/README.rst
@@ -3,15 +3,7 @@ Docker Deployment
 
 .. TODO
 
-    Is there some way to simplify these instructions? It's still a long, error-prone process! Do we need another setup script that follows these directions and prompts the user for input (do you want to run the server or do development? What is your github userid? Etc.)? Such a script would be able to take care of the first 7 steps, which would be good progress. But, we'd need os-dependent stuff for the first bits (install WSL/Ubuntu on Windows); can we assume Python 3 on Mac?
-
-    Enable the script to download and install the Docker Desktop if necessary.
-
-        Blocker: how to determine if the Docker Desktop is installed in Windows? The discussion on `Poweruser <https://superuser.com/questions/68611/get-list-of-installed-applications-from-windows-command-line>`__ on WMIC didn't work for me.
-
-        Blocker: how to determine if OS X is running on x86 or M1? Python doesn't provide accurate info on older x86 versions of the code (see `SO <https://stackoverflow.com/questions/66842004/get-the-processor-type-using-python-for-apple-m1-processor-gives-me-an-intel-pro>`__). Using a specific library just to detect this seems like more trouble than it's worth.
-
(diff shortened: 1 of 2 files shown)


bjones1:
This is a larger PR, though most of the changes are in the docs. Hopefully, it's the last major fix to the Docker stuff for a while.

bnmnetp:
Is this going to force me and others with working setups to change our directory structure?  I can't tell.

bjones1:
No (mostly):
- After this PR, `runestone init` will clone repos in `rsdocker/` instead of the current directory.
- On OS X, it will now insist you run `docker-tools` from a virtual environment.

bnmnetp:
Why will it insist on a VE for macOS?  Not linux?  Not Windows?  That seems like a weird distinction, further just a few weeks ago **nothing** worked if I tried to use a VE now I'm going to have to.

bjones1:
OS X weirdness / a workaround. For some reason, on OS X running `pip install -e docker` to install `rsmanage` means that `rsmanage` isn't on the $PATH. It is for Windows and Linux. However, putting this in a venv on OS X makes these commands work (after activating the venv, which does the $PATH updates needed). I have no idea why OS X does this. The only thing in the venv is `click` and the `rsmanage` and `docker-tools` scripts. The venv is stored in a separate directory from the Docker Poetry environment and doesn't interact with it.

bnmnetp:
I guess for me because I installed python using homebrew both docker-tools and rsmanage are installed in the /opt/homebrew/bin path which is added to one's path by homebrew.

I'm concerned about forcing someone to have a particular virtual environment as I have many different VE
s that I move between all of which allow me to run poetry, docker-tools, and rsmanage.

bnmnetp:
I don't mean to make it more complicated, but I wonder if the confusion isn't from people on older macos using the delivered version of python.  In reality I think we should be telling people to install homebrew and then install a version of python with homebrew.  ?  homebrew is also a good way to install docker.  `brew install --cask docker` gives you the same docker desktop app as downloading a dmg.

bnmnetp:
OK, I'm going to make a test user on my mac and see if I can figure out what is going on.

bjones1:
Both David and Barb had new new M1 Macs. We never touched Python 2; I thought OS X didn't ship with Python 3 at all.

I fully agree with using homebrew to install Docker -- that's now in [this PR](https://github.com/RunestoneInteractive/RunestoneServer/pull/1977/files#diff-048861daf766efd50311c5aba6b2586ccfe82698d3dd9470786b0826004a2b6dR45).

Thanks for Mac help. At some point, I'll rent another macincloud, but I have so little need for that other than the occasional Docker stuff.

bjones1:
OK, I finally integrated these updates. I didn't realize this wasn't a draft PR and force-pushed a rebase and some fixes.

bjones1:
OK, finally wrapping this up. Ready for (re-)review.

PR #1980, opened 2022-07-12
Clean: Remove unused poetry-fix script, clean up docs.

bjones1:


bnmnetp (commented):
I'm confused - the title says that you are removing the script, yet the documenation here says that you **must** run this script.

bjones1:
Sorry for the confusion. I mean, Poetry no longer creates a [script](https://python-poetry.org/docs/pyproject/#scripts) named `poetry-fix` that can be run directly from the CLI; instead, it's run using `python3 runestone/poetry_fix.py`.

bnmnetp:
OK, but now is this a script that **I** must run?  I don' t think I've ever had to run that script manually before.  If so must I do it every time before I run a `poetry update`. or just one time?

bjones1:
Usually, `docker-tools` runs this automatically for you, so you don't notice. If you get complaints about missing `bookserver-dev` or `runestone-dev` paths, running this manually might fix it. There's no change to the ways this works; the PR merely removes something I never ended up using anyway.

bnmnetp:
OK, maybe the docs should say that?  Or maybe it doesn't hurt if people run this when they don't need to?

bjones1:
I'll update the docs to add this. It never hurts to run the program, but it needs to be run with the (Poetry-like) correct flags: `python runestone/poetry_fix.py` for development, but `python runestone/poetry_fix.py --no-dev` for production.

I'm hoping/assuming that most people never need to run this manually, since docker-tools takes care of it...

bjones1:
Where do you want docs? Currently, `pyproject.toml` has the following at the top of the file:

> **Important**: Before running ``poetry install`` or ``poetry update``, run ``python runestone/poetry_fix.py`` so that development dependencies will be installed correctly.

What else would be helpful? Where would you like to put it?

bjones1:
Based on today's hang-out, I updated the docs -- how does that look?

bjones1:
Thanks for the helpful discussion! Getting docs improved is always good.

Issue #1955, opened 2022-06-27
Alphabetizing Names in Instructor Dashboard


kmcdonnell2:
**Describe the solution you'd like**
In the "Add TA" tab for Instructors, it would be nice for the names to be sorted in the "Remove a Current Instructor" box, as they are in the "Add Instructor" box.

Of course the workaround is just to search for it.

PR #1990, opened 2022-08-02
Sorted the list of instructor in-place inside the admin.html file

moisedk:
solved issue #1955  By sorting the instructors' and Ta's names in place based on their last names. 
I assumed that each name is made of a first name and the last name, separated by a space. The way the `instructors` dict is built inside admin.py made me assume this should be valid.


bnmnetp:
closes #1955

PR #1995, opened 2022-08-15
Fix: Set up Alembic to allow migrations by stamping the db on creation.

bjones1:
Fixes [#106](https://github.com/RunestoneInteractive/BookServer/issues/106).

### DIFF
--- a/docker/docker_tools.py
+++ b/docker/docker_tools.py
@@ -1001,6 +1001,7 @@ def _build_phase_2_core(
         xqt("rm -f $RUNESTONE_PATH/databases/*")
         print("Populating database...")
         xqt("rsmanage initdb --force", cwd=env.WEB2PY_PATH)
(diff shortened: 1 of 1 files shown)


docker/docker_tools.py
```diff
@@ -1001,6 +1001,7 @@ def _build_phase_2_core(
         xqt("rm -f $RUNESTONE_PATH/databases/*")
         print("Populating database...")
         xqt("rsmanage initdb --force", cwd=env.WEB2PY_PATH)
+        xqt("alembic stamp head", cwd=env.BOOK_SERVER_PATH)
```
bnmnetp, line 1004: Does this get run with the poetry shell active?  I'm guessing yes but thought I would ask.

bjones1: Good question -- I had to look at the code to double check. Yes, the poetry venv is active when this runs (see https://github.com/RunestoneInteractive/RunestoneServer/blob/master/docker/docker_tools.py#L875).

PR #1998, opened 2022-08-17
Clean: Move pretext dependency into alphabetized order.

bjones1:


pyproject.toml
```diff
@@ -39,6 +39,7 @@ oauth2 = ">=1.9"
 pandas = "^1.0.0"
 pathlib2 = "^2.0.0"
 Paver = ">= 1.2.4"
+pretext = "^1.0.0"
```
bnmnetp, line 42: I know you like to keep things in order, but its a constant battle against the `poetry add` command thats gonna put things where it will.

PR #2025, opened 2022-11-06
Fix: Perform migrations using Alembic on the bookserver for tables...

bjones1:
shared by both this server and the bookserver. See also https://github.com/RunestoneInteractive/BookServer/pull/112.


bjones1:
Tests fail, but not because of this PR.

PR #2040, opened 2023-01-27
Add book rebuild docs

ezrichards:
Add a bit of info on how to manually rebuild the book (this was an issue we had to work through in the Discord).

I'll also look into adding rsmanage features to ensure things are cloned into the right folder. (i.e. a custom fork of thinkcspy)

### DIFF
--- a/rsmanage/rsmanage.py
+++ b/rsmanage/rsmanage.py
@@ -386,22 +386,16 @@ def build(config, clone, ptx, gen, manifest, course):
         if os.path.exists(course):
             click.echo("Book repo already cloned, skipping")
         else:
-            res = subprocess.call("git clone {}".format(clone), shell=True)
+            res = subprocess.call("git clone {} {}".format(clone, str(clone).split("/")[4].split(".")[0]), shell=True)
             if res != 0:
                 click.echo(
                     "Cloning the repository failed, please check the URL and try again"
                 )
                 exit(1)
 
-    # if course name is different from book name, clone to the right folder
-    if not os.path.isdir(course):
-        course = str(clone).split("/")[4]
-
-    #proj_dir = os.path.basename(repo).replace(".git", "")
+    # proj_dir = os.path.basename(repo).replace(".git", "")
     click.echo("Switching to book dir {}".format(course))
-
     os.chdir(course)
-
     if ptx:
         res = _build_ptx_book(config, gen, manifest, course)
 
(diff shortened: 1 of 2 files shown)


ezrichards:
I'm not sure if this is a bug, but it seems like the order of flags actually matters; i.e. this does **not** work for me:
`rsmanage build thinkcspy --clone https://github.com/RunestoneInteractive/thinkcspy.git`
..whereas this does:
`rsmanage build --clone https://github.com/RunestoneInteractive/thinkcspy.git thinkcspy`

ezrichards:
I also added an rsmanage fix; for instance, when running:
`rsmanage build --clone https://github.com/ezrichards/thinkcspy CustomCourse`

..it wouldn't find the directory `CustomCourse`. I added some code for it to redirect to the directory located after the user portion of the URL; `https://github.com/ezrichards/` -> `thinkcspy`

rsmanage/rsmanage.py
```diff
@@ -393,9 +393,15 @@ def build(config, clone, ptx, gen, manifest, course):
                 )
                 exit(1)
 
-    # proj_dir = os.path.basename(repo).replace(".git", "")
+    # if course name is different from book name, clone to the right folder
+    if not os.path.isdir(course):
```
bnmnetp, line 397: This is a problem... The course name has already been checked to be sure it is in the database.  So you can't just change it now to match what was on the github url.

It would be better to change line 389 to have the clone command always clone the url into the name of the course.

ezrichards: Actually, come to think of it, wouldn't this make the course name flag obsolete (as it just copies to the folder anyways)?

bnmnetp: No, because we need to verify that the course is in the database.

ezrichards: Alright, I've changed it to clone into the course name directory, however, this causes this error:
`Error: Example and thinkcspy do not match.  Your course name needs to match the project_name in pavement.py
Build Failed, see cli.log for details`

Do I need to change more than just the clone command? I'm starting to wonder if I was just cloning wrong originally.

ezrichards: I think the first method was where I was going with this. Is pavement.py edited anywhere to begin with or do I need to write new code for that? I've written some code for cloning to the correct directory.

Issue #2048, opened 2023-02-27
Build cleanups


bnmnetp:
@bjones1 

When you do a docker build the default is --single, which seems to be a fairly common thing for people kicking the tires to do.

I fixed the problem with running alembic by conditioning that on being in development mode.

But I noticed that the default `docker/.env.prototype` we set SERVER_CONFIG to development unconditionally.

I've added a bit to the `docker/README.rst` file, but I wonder if we can make that smarter?

It was this mismatch that ended up causing the bookserver to fail to start because it tried to change the the BookServer folder thinking it was in development mode.

bjones1:
Thanks for tracking this down! I looked a bit, but another project claimed all my time this weekend.

Given that most people will probably want to run the server (production mode, which is after all the default when running `docker-tools build`), let's edit `.env.prototype` to set `SERVER_CONFIG=production`. Then, update the Docker `readme.rst` instructions to tell developers to change this.

bjones1:
In terms of making it smarter, we currently `cp docker/.env.prototype .env` if `.env` doesn't exist. Instead of the copy, we could auto-edit the `SERVER_CONFIG`. If we wanted to be more clever, we could also run `md5sum`, and re-update/edit this if the file hasn't been hand-edited (the md5 checksum didn't change).

PR #2049, opened 2023-02-27
Improve docker-tools for `build --single`

bjones1:
This implements ideas in #2048. Not much testing yet.

Fix: 
- auto-edit .env based on the build type.
- Links development-only files only in dev mode.

Docs: 
- Improve docs on .env.prototype

### DIFF
--- a/docker/.env.prototype
+++ b/docker/.env.prototype
@@ -7,7 +7,7 @@
 RUNESTONE_HOST=localhost
 
 # Select a configuration for the instructor-facing server (the Runestone server) and the student-facing server (BookServer). Valid values are ``test``, ``development``, and ``production``.
-SERVER_CONFIG=development
+SERVER_CONFIG=${SERVER_CONFIG}
 
 # For production, change the credentials for the DB to something more secure.
 # This should be done prior to first running ``docker/docker_tools.py up``.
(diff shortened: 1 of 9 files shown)


bjones1:
@bnmnetp. OK works for me after testing. Test failures may be OK; the tests need to be re-run.

bnmnetp:
tracking back a little further it is trying to run rsmanage but 
```
from rsmanage import cli
runestoneserver-runestone-1  | ImportError: cannot import name 'cli' from 'rsmanage'
```

bnmnetp:
This is so weird, I can't even find anyplace in our code where we try to import cli from rsmanage.  Running the tests locally on my dev build and all is fine, so it seems to be something in the setup for the tests.

Any ideas???

bjones1:
I'm guessing this is something new/broken with Poetry 1.4.0. In `rsmanage/setup.py`, we have:

```
    entry_points="""
        [console_scripts]
        rsmanage=rsmanage:cli
    """,
```
which is probably where the import of `cli` comes from. I'll test and see what happens.

bjones1:
OK, testing a fix now...

bnmnetp:
looks like it somehow lost the docker-tools test command...

bnmnetp:
I'm seeing this message:
```
print("Note: this must be an initial install; additional commands missing.")
```

So something is going wrong with the import from `docker_tools_misc`

bnmnetp:
my guess is that we now need to import `tests.ci_utils` not sure where this is coming from... poetry?? New setuptools?

bnmnetp:
Some real backward incompatibility in poetry 1.4?  I see it was just released on the 2/27.  I had rebuilt locally a few days before. Looking at the changelog I didn't see see anything to indicate such major changes.

bjones1:
That makes sense to me -- these fixes remove a lot of Poetry bug fix/workaround stuff, so I think it's a good step forward. Let's merge this, then you merge your 3.10 branch. Hopefully, that will get us back to all tests passing again!

bjones1:
Do you want to merge this, or shall I?

bnmnetp:
I'm on it.

PR #2054, opened 2023-03-21
Reduce worker count on development deploys

ascholerChemeketa:
Launching large numbers of workers causes issues on high core count machines when developing in WSL


bjones1:
@bnmnetp. this looks good to me! I'll let you merge.

PR #2057, opened 2023-03-30
Pull Request for the CodeChat version of our Initial troubleshooting and debugging.

DanielShort137:
We are attempting to modify and run the debugger from command line using the pdb python library.  The Code chat editor was used to create code and doc-blocks within the test_server.py file.  We also added the links to websites for additional debugger information, since we thought it might help the next group to look over this code.  There are many HTML tags that were added from codechat parser.  This request is for both Josh and Daniel.


bnmnetp:
Hi @DanielShort137 I'm not really clear what you are trying to accomplish with this PR, maybe you can say a bit more about your goals.  If you are just trying to run and debug failing tests you can use the `--pdb` option with pytest and it will automatically start up the debugger when a test fails.

@bjones1 -- Does that look like correct behavior from CodeChat ? I didn't think we would see html tags in the comments in the code...

DanielShort137:
Hi @bnmnetp  this was an assignment given by @bjones1 for his Literate Programming Class.  If I misunderstood the assignment sorry.  I will follow up and ensure that I did not create this request in error.

